PR #8, opened 2021-08-24
Wave equation ssprk2

amartinhuertas:
Hi @cbladwell, @davelee2804,

in this PR I have developed a first implementation (I want to subject it to review from your side) of the SSPRK2 (Strong Stability Preserving explicit Runge Kutta 2nd order) time integration method for the linear wave equation. In this implementation, the time integration vector updates are performed using lower level functions (still of Julia's high level API though) than the ones in your implementation. **The goal is to reduce to the minimum the amount of dynamic memory allocation during the time integration loop.** Please note that I have also tried to maximize the usage of Gridap's high level API. Sure that it can be improved, but this is what I have in mind of a reasonable simulation driver.

I think the implementation is correct. With 10x10 quads/panel (analytical geometrical mapping), T=2*PI, g=H=1, 2000 time steps, RT0-DG0, the potential to kinetic energy balance looks like as in Chris's slides:

kinetic_to_potential_energy_balance

Please note the following: 
 1. The `assemble_matrix` Gridap's high level API function let us compute a matrix by FE assembly in a single call out of the bilinear form and FE spaces. See [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L115). There is also an `assemble_vector` counterpart. 
 2. The time integration loop is fully performed using linear algebra data structures (i.e., vectors and matrices). We do not actually need to create `FEFunction` instances along the process. We create them at the beginning, and extract pointers out of them (via `get_free_dof_values`) that we use along the loop. 
 3. The LU factorization of the mass matrices is computed once (see, e.g., [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L117)) before the time integration loop, and reused over and over again at each iteration of the loop (see, e.g., [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L68)). If we used `solve` with an `AffineFEOperator` instance (I think) we would be computing the lu factorization at each time step!!!
 4.  The non-diagonal component matrices div(v)*p and q*div(u) are assembled only once at the beginning using reference element integration (see, e.g., [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L122)) and re-used over and over again at each iteraiton of the loop (see [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L67))
5. Vector updates that solely involve vectors are performed using dot notation (see, e.g., [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L163)) so that they are performed with a single vectorized loop and no temporary dynamic memory allocation. (You are already aware of this, we have talk about this before, anyway it worths remarking.)
6. I am using FE interpolation (i.e. evaluation of DoFs) instead of L2 projection to inject the initial condition to the FE space (see, e.g., [here](https://github.com/gridapapps/GridapGeosciences.jl/blob/3e92f4742fbe4e2129c94ace68cb2c886ad6fcb1/test/WaveEquationCubedSphereTests.jl#L127). Note that interpolation is also a valid type of projection in itself.
7. In many places (although not in all places), there is no need to explicitly use the `CellField` constructor with analytical functions,  Gridap calls it under the hood for you.  

Let me know your thoughs about the code. We can discuss it at any moment if you like.

### DIFF
--- a/test/WaveEquationCubedSphereTests.jl
+++ b/test/WaveEquationCubedSphereTests.jl
@@ -1,0 +1,197 @@
+module WaveEquationCubedSphereTests
+
+using Test
+using Gridap
+using GridapGeosciences
+using Plots
+using LinearAlgebra
+using WriteVTK
+
+# Initial depth
+function h₀(xyz)
+  x,y,z = xyz
+  xᵢ,yᵢ,zᵢ = (1,0,0)
+  1.0 + 0.0001*exp(-5*((xᵢ-x)^2+(yᵢ-y)^2+(zᵢ-z)^2))
+end
+
+# Initial velocity
+u₀(xyz) = zero(xyz)
+
+"""
+  Kinetic energy
+"""
+function Eₖ(uh,H,dΩ)
+  0.5*H*sum(∫(uh⋅uh)dΩ)
+end
+
+"""
+  Potential energy
+"""
+function Eₚ(hh,g,dΩ)
+  0.5*g*sum(∫(hh*hh)dΩ)
+end
+
+"""
+  Total energy
+"""
+function Eₜ(uh,H,hh,g,dΩ)
+  Eₖ(uh,H,dΩ)+Eₚ(hh,g,dΩ)
+end
+
+"""
+  Kinetic to potential
+"""
+function compute_kin_to_pot!(w,unv,divvh,hnv)
+  mul!(w,divvh,hnv)
+  unv⋅w
+end
+
+"""
+  Potential to kinetic
+"""
+function compute_pot_to_kin!(w,hnv,qdivu,unv)
+   mul!(w,qdivu,unv)
+   hnv⋅w
+end
+
+
+function new_vtk_step(Ω,file,hn,un)
+  createvtk(Ω,
+            file,
+            cellfields=["hn"=>hn, "un"=>un],
+            nsubcells=4)
+end
+
+function _ssrk2_update!(res,MM,MMchol,DIV,α,a,b)
+  mul!(res, MM, a)            # res <- MM*a
+  mul!(res, DIV, b, α, 1.0)   # res <- 1.0*res + α*DIV*b
+  ldiv!(MMchol,res)           # res <- inv(MM)*res
+end
+
+function generate_energy_plots(outdir,N,ke,pe,kin_to_pot,pot_to_kin)
+  plot(1:N,
+       [ke,pe],
+       label=["Ek (kinetic)" "Ep (potential)"],
+       xlabel="Step",ylabel="Energy")
+  savefig(joinpath(outdir,"energy.png"))
+  plot(1:N,
+       [kin_to_pot,-pot_to_kin,kin_to_pot-pot_to_kin],
+       label=["hⁿ∇⋅uⁿ" "uⁿ⋅∇(hⁿ)" "Balance"],
+       xlabel="Step",ylabel="Kinetic to potential energy balance",legend = :outertopleft)
+  savefig(joinpath(outdir,"kinetic_to_potential_energy_balance.png"))
+end
+
+"""
+  Solves the wave equation using a 2nd order
+  Strong-Stability-Preserving Runge-Kutta
+  explicit time integration scheme (SSRK2)
+
+  g : acceleration due to gravity
+  H : (constant) reference layer depth
+  T : [0,T] simulation interval
+  N : number of time subintervals
+"""
+function solve_wave_equation_ssrk2(
+      model,order,degree,g,H,T,N;
+      write_results=false,
+      out_dir="wave_eq_ncells_$(num_cells(model))_order_$(order)_ssrk2",
+      out_period=N/10)
+
+  RT=ReferenceFE(raviart_thomas,Float64,order)
+  DG=ReferenceFE(lagrangian,Float64,order)
+  V = FESpace(model,RT; conformity=:Hdiv)
+  Q = FESpace(model,DG; conformity=:L2)
+
+  U = TrialFESpace(V)
+  P = TrialFESpace(Q)
+
+  Ω  = Triangulation(model)
+  dΩ = Measure(Ω,degree)
+  dω = Measure(Ω,degree,ReferenceDomain())
+
+  # Build mass matrix in RT and DG spaces
+  # and their sparse Cholesky factors
+  amm(u,v) = ∫(v⋅u)dΩ
+  RTMM=assemble_matrix(amm,U,V)
+  L2MM=assemble_matrix(amm,P,Q)
+  RTMMchol=lu(RTMM)
+  L2MMchol=lu(L2MM)
+
+  # Build g*DIV(v)*h and H*q*DIV(u)
+  ad(h,v) = ∫(DIV(v)*h)dω
+  divvh=assemble_matrix(ad,P,V)
+  adt(u,q) = ∫(q*DIV(u))dω
+  qdivu=assemble_matrix(adt,U,Q)
+
+  # Interpolate initial condition into FE spaces
+  hn=interpolate_everywhere(h₀,P); hnv=get_free_dof_values(hn)
+  un=interpolate_everywhere(u₀,U); unv=get_free_dof_values(un)
+  if (write_results)
+    rm(out_dir,force=true,recursive=true)
+    mkdir(out_dir)
+  end
+  pvdfile=joinpath(out_dir,"wave_eq_ncells_$(num_cells(model))_order_$(order)_ssrk2")
+  paraview_collection(pvdfile) do pvd
+    # Allocate work space vectors
+    h1v = similar(get_free_dof_values(hn))
+    h2v = similar(h1v)
+    u1v = similar(get_free_dof_values(un))
+    u2v = similar(u1v)
+    if (write_results)
+      ke  = Vector{Float64}(undef,N)
+      pe  = Vector{Float64}(undef,N)
+      kin_to_pot = Vector{Float64}(undef,N)
+      pot_to_kin = Vector{Float64}(undef,N)
+    end
+    dt  = T/N
+    dtg = dt*g
+    dtH = dt*H
+    for step=1:N
+       # 1st step
+       # inv(L2MM)*(L2MM*hnv - dt*Hqdivu*unv)
+       # inv(RTMM)*(RTMM*unv + dt*gdivvh*hnv)
+       @time _ssrk2_update!(h1v, L2MM, L2MMchol, qdivu,-dtH, hnv, unv)
+       @time _ssrk2_update!(u1v, RTMM, RTMMchol, divvh, dtg, unv, hnv)
+
+       # 2nd step
+       # inv(L2MM)*(L2MM*h1v - dt*Hqdivu*u1v)
+       # inv(RTMM)*(RTMM*u1v + dt*gdivvh*h1v)
+       @time _ssrk2_update!(h2v, L2MM, L2MMchol, qdivu, -dtH, h1v, u1v)
+       @time _ssrk2_update!(u2v, RTMM, RTMMchol, divvh,  dtg, u1v, h1v)
+
+       # Averaging steps
+       hnv .= 0.5 .* ( hnv .+ h2v )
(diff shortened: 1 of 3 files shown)


davelee2804:
Great work @amartinhuertas . I assume from your power plot that gaussian bell makes one full revolution at step 2000 (time=2\pi), which is what we expect. Have you visually confirmed this? (I'd be 99% sure its correct, but just to be safe).

Just a couple of minor comments:

1) Your diagnostic functions, for computing energy and kinetic to potential power exchanges look really nice, but I would think these should be refactored into a separate file so they can be re-used by other configurations

2) To be safe I would also check the global mass conservation (again, I'd be 99% sure this is fine)

3) The naming of RTMMchol, L2MMchol, are you happy with this? I assume they are called this due to a Cholesky factorisation, but however they are computed, they are really inverses yes? I would personally name them RTMMinv, L2MMinv, but it's up to you,

Two final comments:

1) Regarding the high level DIV operator, do you also have a high level GRAD-PERP? If so I will start experimenting with these...

2) After your helpful comments yesterday I'm sure my implementation of the APVM term in the mass_flux branch is wrong, and this should instead be using your \nabla on the manifold operator. Just be aware of this if you are looking at this term.....

amartinhuertas:
Hi @davelee2804 ... thanks for your comments! Let me answer them step by step! 

> Great work @amartinhuertas . I assume from your power plot that gaussian bell makes one full revolution at step 2000 (time=2\pi), which is what we expect. Have you visually confirmed this? (I'd be 99% sure its correct, but just to be safe).

Yes, I did. Besides, I just added to the README.md file several animations with different mesh resolutions and polynomial order. It takes a while to load the first time, once loaded, I guess that it is kept in the cache of the browser. You can then press reload, and then you should see all simulations synchronized. 

https://github.com/gridapapps/GridapGeosciences.jl/tree/wave_equation_ssprk2

> Your diagnostic functions, for computing energy and kinetic to potential power exchanges look really nice, but I would think these should be refactored into a separate file so they can be re-used by other configurations

Do you think they are general enough to be used in other driver programs as well? (not much experience)

> To be safe I would also check the global mass conservation (again, I'd be 99% sure this is fine)

Ok, this could be a task for @cbladwell ... what do you think @cbladwell ? 

> The naming of RTMMchol, L2MMchol, are you happy with this? I assume they are called this due to a Cholesky factorisation, but however they are computed, they are really inverses yes? I would personally name them RTMMinv, L2MMinv, but it's up to you,

They are not inverses. AFAIK, Inverses in computer programs are rarely computed (because of numerical stability issues). Instead, they store the L,U factors resulting from UMFPACK factorization. These factors are used later on in `ldiv!` to solve the linear system using forward/backward substition. I would have like to use `cholesky` Julia function, instead of `lu` function. However, it turns out that `ldiv!` is not available for `cholesky`, so that I had to resort to general LU factorization. I guess this is not efficient, because the matrix is SPD. 

> Two final comments:

Let us clarify this in slack better.

cbladwell:
H @amartinhuertas thanks for sharing this implementation. I've reviewed and run it. I think this approach should be straightforward to implement for other time integrators. 

> To be safe I would also check the global mass conservation (again, I'd be 99% sure this is fine)

I've just checked - I think the mass conservation for the `Geosciences` impelementation and it is `8.01136934569513e-13` (for reference for the 12x12 mesh model in `Geofem` the mass conservation is 1.255884285455977e-12 )

> Your diagnostic functions, for computing energy and kinetic to potential power exchanges look really nice, but I would think these should be refactored into a separate file so they can be re-used by other configurations

> Do you think they are general enough to be used in other driver programs as well? (not much experience)

The diagnostics are great. I think it would be good to have diagnostics that can be used for different integrators. I think they will have to be specific to the PDE we are solving

cbladwell:
H @amartinhuertas and @davelee2804 thanks for sharing this implementation. I've reviewed and run it. I think this approach should be straightforward to implement for other time integrators. 

> To be safe I would also check the global mass conservation (again, I'd be 99% sure this is fine)

I've just checked - I think the mass conservation for the `Geosciences` impelementation and it is `8.01136934569513e-13` (for reference for the 12x12 mesh model in `Geofem` the mass conservation is 1.255884285455977e-12 )

> Your diagnostic functions, for computing energy and kinetic to potential power exchanges look really nice, but I would think these should be refactored into a separate file so they can be re-used by other configurations

> Do you think they are general enough to be used in other driver programs as well? (not much experience)

The diagnostics are great. I think it would be good to have diagnostics that can be used for different integrators. I think they will have to be specific to the PDE we are solving

amartinhuertas:
> I've just checked - I think the mass conservation for the Geosciences impelementation and it is 8.01136934569513e-13 (for reference for the 12x12 mesh model in Geofem the mass conservation is 1.255884285455977e-12 )

Ok. This difference might be explained by the analytical versus bilinear map (if you were using the analytical map in Geosciences). You can also leverage the bilinear map in GridapGeosciences.jl. You just have to call the constructor of the mesh as `CubedSphereDiscreteModel(12,1)` instead of `CubedSphereDiscreteModel(12)`

davelee2804:
Thanks for the comments @amartinhuertas , @cbladwell .

1) mass conservation: thanks for checking @cbladwell . Anything < 1.0E--12 should be OK I think :). With some integrators there is a machine precision drift in the mass conservation with each time step, which might be the case for the SSPRK2 integrator. Did you plot it with time, is there a linear trend?

2) diagnostics: I think these are general enough to be put in a common location. The power exchange will always be `<something in L2,DIV(something in RT)>`, the potential energy is the same for shallow water and wave equations. The kinetic energy will be different, but that is fine :)

3) Thanks for the confirmation about the LU factors @amartinhuertas , sorry I should have realised that (I was surprised that you were carrying around exact matrix inverses :) )

I replied to the DIV GRAD-PERP question in slack...

amartinhuertas:
> mass conservation: thanks for checking @cbladwell . Anything < 1.0E--12 should be OK I think :). With some integrators there is a machine precision drift in the mass conservation with each time step, which might be the case for the SSPRK2 integrator. Did you plot it with time, is there a linear trend?

@cbladwell ... when you have the code ready with the mass conservation evaluation, can you please push it? Thanks! Could you compare the computational times? 

> diagnostics: I think these are general enough to be put in a common location. The power exchange will always be <something in L2,DIV(something in RT)>, the potential energy is the same for shallow water and wave equations. The kinetic energy will be different, but that is fine :)

Ok. How is the kinetic energy defined for the shallow water Eqs ? 

> Thanks for the confirmation about the LU factors @amartinhuertas , sorry I should have realised that (I was surprised that you were carrying around exact matrix inverses :) )

No problem. As I said, I think we should try to see how we can exploit symmetry. why `ldiv!` does not work with `cholesky`? perhaps we can ask at julia slack channel.

cbladwell:
> @cbladwell ... when you have the code ready with the mass conservation evaluation, can you please push it? Could you compare the computational times?

I've push the mass diagnostic now. The computational times are

```
Geosciences: 46.574903 seconds (99.52 M allocations: 6.696 GiB, 3.94% gc time)
```
```
Geofem: 144.50280 seconds (564.13 M allocations: 43.185 GiB, 5.77% gc time)
```

> With some integrators there is a machine precision drift in the mass conservation with each time step, which might be the case for the SSPRK2 integrator. Did you plot it with time, is there a linear trend?

@davelee2804 the drift seems linear with slightly better mass conservation in the new `Geosciences` (see below). 

mass_cons_drift

amartinhuertas:
> I've push the mass diagnostic now.

Ok, thanks very much! I have some comments though ... I will add them to the commit. 

> The computational times are

Not bad. 3x reduction of computational times and  6.5x reduction in the total memory allocated. 

> @davelee2804 the drift seems linear with slightly better mass conservation in the new Geosciences implementation (see below).

Did you run both with the bilinear map for the approximation of the sphere? Or ... one with the analytical map and the other with the bilinear map?

wave_eq_ncells_600_order_0_ssrk2/wave_eq_ncells_600_order_0_ssrk2.pvd
```diff
@@ -0,0 +1,206 @@
+<?xml version="1.0" encoding="utf-8"?>
```
amartinhuertas, line 1: Remove simulation results from the repo. (I guess that you accidentally pushed them).

amartinhuertas: Add vtu, pvd and jld in gitignore file so that we avoid this in the future.

test/WaveEquationCubedSphereTests.jl
```diff
@@ -54,6 +55,14 @@ function compute_pot_to_kin!(w,hnv,qdivu,unv)
    hnv⋅w
 end
 
+"""
+Compute the total mass
+"""
+function compute_mass(L2MM,hh)
+  (L2MM*hh)⋅ones(length(hh))
```
amartinhuertas, line 62: This function allocates memory dynamically at two different points:

1) `ones(length(hh))` allocates a whole vector of ones. Note that the dot product against a vector of ones is equivalent to summing up all the entries of the other vector. 
2) `L2MM*hh` creates a temporary output array with the matrix vector product. Use `mul!` and an input workspace vector as we have done in other functions in this script.

test/WaveEquationCubedSphereTests.jl
```diff
@@ -142,6 +151,7 @@ function solve_wave_equation_ssrk2(
       pe  = Vector{Float64}(undef,N)
       kin_to_pot = Vector{Float64}(undef,N)
       pot_to_kin = Vector{Float64}(undef,N)
+      mass = Vector{Float64}(undef,N)
```
amartinhuertas, line 149: This memory allocation is ok. We allocate before the time integration loop. Not at each iteration, like in the previous function.

amartinhuertas: Besides, as we know the size of the vectors beforehand, we can allocate to given size, instead of initializing to the empty vector and then pushing entries to the end (which again, requires frequent memory allocation)

davelee2804:
Thanks for sharing the plot @cbladwell . I've seen this sort of thing before, (see fig. 14 here: https://www.sciencedirect.com/science/article/pii/S0021999119306989) I suspect that going to 3rd order RK would get rid of the linear mass conservation error, but I'm not sure if that is worth looking into. I think that happens is the linear intermediate approximation in the RK2 time stepping leads to a machine precision overshoot in the mass at each time step. I don't think it is worth worrying about. I don't think the semi-implicit time stepping will suffer from this....

cbladwell:
> I suspect that going to 3rd order RK would get rid of the linear mass conservation error, but I'm not sure if that is worth looking into. I think that happens is the linear intermediate approximation in the RK2 time stepping leads to a machine precision overshoot in the mass at each time step. I don't think it is worth worrying about. I don't think the semi-implicit time stepping will suffer from this....

That's interesting. The mass conservation for the semi implicit is `3.552713678800501e-15` and there isn't any drift in the time plot. It might be interesting to look into higher order RK schemes with this approach

davelee2804:
Hey @cbladwell , thanks for the confirmation. Yes it is a curious thing that some explicit schemes suffer from this subtle problem.

wave_eq_ncells_600_order_0_ssrk2/wave_eq_ncells_600_order_0_ssrk2.pvd
amartinhuertas: Done!

PR #9, opened 2021-08-25
Arbitrary radius CubedSphereDiscreteModels + Earth constants

amartinhuertas:


davelee2804:
This is a really great develpment @amartinhuertas  - all the test cases I know of are configured in physical units, and while its not too hard to non-dimensionalise them to a unit sphere, it is easier still to just expand the sphere :)

amartinhuertas:
> This is a really great develpment @amartinhuertas - all the test cases I know of are configured in physical units, and while its not too hard to non-dimensionalise them to a unit sphere, it is easier still to just expand the sphere :)

Yeah. Note that I have defined a new function-like struct, i.e., instances of this data type are callable. Using this sw pattern I can parametrized the map from the cube surface to the sphere by the radius of the latter.

PR #10, opened 2021-08-26
Theta method full newton nswe solver

amartinhuertas:
Hi @davelee2804 @cbladwell,

in this PR I have written a "skeleton" code (it is just a partial implementation, to be completed/validated) of a NSWE solver using Gridap's high level API. 

* It uses the APVM method for spatial discretization, i.e., Eqs. (70)-(73) in [Rognes et. al](https://gmd.copernicus.org/articles/6/2099/2013/) and the theta method implicit method for time discretization (the fully discrete nonlinear problem to be solved at each time step is in particular avaliable in Eqs. (75)-(78) of that paper). 
* You can see the evaluation of this nonlinear operator (i.e., residual) coded here: https://github.com/gridapapps/GridapGeosciences.jl/commit/88418fe8cdf327412822361733626dd4988056db#diff-3d44a05a1e12bbf492c4d4eea9a151784def9fe3a3c34312c3e9a5f35e106ff6R208
where the first argument is the current evaluation point, and the latter a basis of the monolothic system test functions.

However, some caveats:
 1. I could **NOT** compute the Jacobian of this operator using AD (because of a known issue in Gridap 0.16 that has to be solved; see https://github.com/gridap/Gridap.jl/issues/584 for details) We will need an approximate Jacobian, or to resort to a different discretization, e.g., the one at the end of Section 5.5 of Rognes paper (this discretization is not energy conserving, though). 
 2. The JIT compilation of the code triggered from `res` down the tree takes forever with Julia 1.6.2 (it seems to enter in a infinite loop, although I could not confirm it). With Julia 1.5.4 the issue disappears. Again, I think similar issues have appeared with Julia 1.6+Gridap in similar contexts. We should investigate this further. 

Also I wonder the following:
- Can we actually exploit the DIV+reference integration for this term: `(∇⋅(v))*(0.5*ui⋅ui)` ?
- Is the current (partial) implementation correct? (as I could not validate it yet) 

@davelee2804 @cbladwell perhaps you can contribute and complete it (essentially implement a crude approximation of the Jacobian as @davelee2804 mentioned). What do u think?  I think that this driver shows that we have all the ingredients in Gridap required to solve the NSWEs at the discretization level (e.g., I could evaluate the residual as implemented in `res`).  I am happy with you completing the current method, or to consider another method as well (e.g., one of Cotter's book for the NSWE different from the APVM). 

Please, once you start modifying the driver, can you please open a branch and PR to this branch so that I can help you and review etc. ? Thanks!

## Tasks 

- [x] Complete the exact Jacobian, i.e., complete the `jac` function definition statement. https://github.com/gridapapps/GridapGeosciences.jl/blob/11fadb3caad4585c5e30526d0db58d865ad0bdb6/test/ThetaMethodFullNewtonNSWECubedSphereTests.jl#L171
- [x] Implement/Complete diagnostics (currently commented out from WaveEquationSolver)
- [ ] Validate the solver against different benchmarks from Williamson et. al. 
- [x] Solve the following question: Can we actually exploit the DIV+reference integration for this term: `(∇⋅(v))*(0.5*ui⋅ui)` ? (see PR https://github.com/gridapapps/GridapGeosciences.jl/pull/12 for more details)
- [x] Investigate issues with AD of the fully coupled system. (no longer needed, we have the manual Jacobian).
- [ ] **UPDATE**: *by a more clever ad-hoc implementation of the Backward Euler method I could reduce first JIT compilation times to around 10 mins with Julia 1.5.4. Not sure with Julia 1.6.2., we will see in Github actions.* Investigate JIT compilation issues with Julia 1.6.2. Switch to Julia 1.5.4 in the meantime.


amartinhuertas:
For the records ... trying to solve this discussion with the code + discussion in PR https://github.com/gridapapps/GridapGeosciences.jl/pull/12

amartinhuertas:
> Hey @amartinhuertas . Can I ask you a couple of questions?

Yes, sure. 

> Are \Delta u, \Delta h the difference between the variable at the previous time level, and the value at the current Newton iterate, ie is:

I think that the answer is no.  \Delta u_n and \Delta h_n, with n=0,1,2 etc.  is the difference among the variable at the current (unknown) and previous (known) time levels (i.e., u_n+1-u_n).  In the method I have implemented, \Delta u_n and \Delta h_n  are indeed the sought-after solution to the non-linear problem at the current time step (along with qvort and F). 

In your reasoning, how is u^{k} defined and why? I dont think there is a u^{k} ... (explanation next)

During the nonlinear iteration we want to approximate \Delta u_{n}=u_n+1-u_n starting from \Delta u_{n-1}=u_n-u_{n-1} as initial guess. Thus, we compute \Delta u_n^{1}, \Delta u_n^{2}, .... \Delta u_n^{k}. After convergence, \Delta u_n^{k} becomes \Delta u_n. 

Does this make sense ? 


> If (1) is true, why do \Delta u, \Delta h appear in the subsequent terms in lines 158 - 162 ? I would assume these terms, ui(\Delta u), hi(\Delta h) should be

I think (1) is false. Thus, I wont answer to this one. If am wrong, then we can reconsider this second point. 


> Should line 171 be:
\int (s*qvort*hi(dh) + ...)

what is `dh` ?

davelee2804:
> After convergence, \Delta u_n^{k} becomes \Delta u_n.

Yes, I agree, so after convergence (or termination of the iteration) `u_{n+1} = u_n + \Delta u_n^{k}`

> what is dh ?

I think there is a bug at 
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/test/ThetaMethodFullNewtonNSWECubedSphereTests.jl#L171
and that it should be `dp` instead of `dr` (sorry, I had just assumed it was called `dh`, it is `dp`)

Also, if you are still having trouble with convergence, I would change the nonlinear terms in lines 158 and 162 so that these are given as:
```
0.25*(u_n\cdot u_n + u_{n+1}^k\cdot u_{n+1}^k)
0.5*(h_n*u_n + h_{n+1}^k*u_{n+1}^k)
```
ie: good old fashioned Crank-Nicolson, just to ensure these nonlinearities do not cause any sort of temporal aliasing issues
(just a suggestion)

amartinhuertas:
> Yes, I agree, so after convergence (or termination of the iteration) u_{n+1} = u_n + \Delta u_n^{k}

Correct.

> I think there is a bug at
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/test/ThetaMethodFullNewtonNSWECubedSphereTests.jl#L171
and that it should be dp instead of dr (sorry, I had just assumed it was called dh, it is dp)

Yes, this is definitely a bug. I agree it should be `dp`. Thanks for spotting this! This is the typical error I need additional eyes to look at the code. I did not test it yet ... (takes more than 30 min to JIT compile the residual + jacobian on first invocation, I should be doing something wrong) 

> Also, if you are still having trouble with convergence, I would change the nonlinear terms in lines 158 and 162 so that these are given as:
> 0.25*(u_n\cdot u_n + u_{n+1}^k\cdot u_{n+1}^k)
> 0.5*(h_n*u_n + h_{n+1}^k*u_{n+1}^k)
> ie: good old fashioned Crank-Nicolson, just to ensure these nonlinearities do not cause any sort of temporal aliasing issues
> (just a suggestion)

If am not wrong, theta method with \theta=0.5 is already Crank-Nicolson, isn't ? Note that if you replace the call to ui() hi() hib() in the nonlinear terms by their definition you already get the expressions you are writing. Right?

davelee2804:
> If am not wrong, theta method with \theta=0.5 is already Crank-Nicolson, isn't ? Note that if you replace the call to ui() hi() hib() in the nonlinear terms by their definition you already get the expressions you are writing. Right?

For the linear terms I'm pretty sure they are the same, but for the nonlinear terms I think they are different (perhaps I am wrong). My interpretation of your `ui(\Delta u)` term is that this is equal to:
```
ui(\Delta u) = \theta un + (1-\theta)u_{n+1}^k
```
(same for `hi(\Delta h)`).
So the nonlinear terms (letting `\theta=0.5` for simplicity of explanation) will be:
```
0.25(un\cdot un+ 2*un\cdot u_{n+1}^k + u_{n+1}^k\cdot u_{n+1}^k)
0.25(hn*un + h_{n+1}^k*un + hn*u_{n+1}^k + h_{n+1}^k*u_{n+1}^k)
```

These terms _slightly_ differ from the exact temporal integrals (which can be found in the paper draft, equations 23 and 24 I think from memory). As such there may be overshoots to these values which may interfere with the convergence, that it my concern here.

davelee2804:
...The exact integrals should be neutrally stable. Crank-Nicolson I think is convex (no temporal cross terms), so it will not be neutrally stable, but rather damping. Alternative formulations of the nonlinear cross terms _may_ lead to overshoots (I'm not 100% sure of this)

davelee2804:
Hey @amartinhuertas , amazing work on getting the backward Euler solve converging, well done!!! 
You may already be thinking in this direction, but I'm guessing the safest next step may be straight Crank-Nicolson (including for the nonlinear terms), so that would be:
```
0.25*(u_n\cdot u_n + u_{n+1}^k\cdot u_{n+1}^k)
0.5*(h_n*u_n + h_{n+1}^k*u_{n+1}^k)
```
just to keep out the other nonlinear cross terms and remove additional sources of uncertainty.
Great job again on getting the coupled solver working!

amartinhuertas:
> You may already be thinking in this direction, but I'm guessing the safest next step may be straight Crank-Nicolson (including for the nonlinear terms), so that would be:

I think this can be work for @cbladwell ... what do you think? We can perhaps add a task on the PR description on this.

davelee2804:
Does that work for you @cbladwell ?

davelee2804:
Also, as you have written in the tasks @amartinhuertas , a convergence study at multiple resolutions might be a good idea. For backwards Euler I'm guessing the errors will be first order

PR #15, opened 2021-09-02
Shallow water explicit

davelee2804:
This pull request is to add a second order explicit shallow water integrator and associated Williamson2 analytic test case into GridapGeosciences.

It includes additions to the src/ directory as follows:
1) additional diagnostics for the shallow water equations (mass, vorticity, energy, power)
2) a reference domain grad_perp operator (low level)
3) diagnostic equation to compute the vorticity (low level)
4) diagnostic equation to compute the potential vorticity (mixed low level/high level)
5) explicit integrator (nonlinearly stable, exactly balances energy exchanges in space and time)
6) integration driver function, takes in an integrator as a function pointer and cycles over this for a specified number of time steps

It also includes an analytic test (williamson2), run at multiple spatial and temporal resolutions in order to determine the L2 errors in the velocity and the fluid depth. L2 errors at 3 resolutions using RT1 and supplied in order to test against and verify second order accuracy


src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+
+  # initialise the diagnostics arrays
+  mass = zeros(0)
+  vort = zeros(0)
+  kin  = zeros(0)
+  pot  = zeros(0)
+  pow  = zeros(0)
+
+  # first step, no leap frog integration
+  hm1          = new_field(Q, hn)
```
amartinhuertas, line 117: I think that `new_field`, though correct, is a bit over-engineered. 

You can do the same more concisely as: 

`FEFunction(Q, copy(get_free_dof_values(hn)))`

so that `new_field` is not required.

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+  # hₘ         : fluid depth at previous time level (for leap-frogging the first step)
+  # uₘ         : fluid velocity at previous time level (for leap-frogging the first step)
+  # RTMM       : H(div) mass matrix, ∫β⋅βdΩ, ∀β∈ H(div,Ω)
+  # L2MM       : L² mass matrix, ∫γγdΩ, ∀γ∈ L²(Ω)
+  # dt         : time step
+  # leap_frog  : do leap frog time integration for the first step (boolean)
+  # order      : polynomial order
+  # dΩ         : measure of the elements
+
+  u = get_trial_fe_basis(U)
```
amartinhuertas, line 29: Why do you need `u` `v` `p` and `q` ?

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+  # short cut routine to initialise a new FEFunction 
+  # A: test function
+  # a: dofs to copy into the FEFunction
+  a_dof  = Gridap.FESpaces.get_free_dof_values(a)
+  b_dof  = similar(a_dof)
+  b_dof .= a_dof
+  b      = FEFunction(A, b_dof)
+end
+
+function shallow_water_time_stepper(model, order, Ω, dΩ, dω, qₖ, wₖ, f, g, hn, un, dt, nstep, dump_freq, τ, P, Q, U, V, R, S, method)
```
amartinhuertas, line 100: why do you need `qₖ, wₖ` ?

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+  # short cut routine to assemble a domain contribution into a vector
+  # A:  test function
+  # B:  trial function
+  # dc: domain contribution
+  data  = Gridap.FESpaces.collect_cell_vector(B, dc)
+  assem = SparseMatrixAssembler(A, B)
+  rhs   = assemble_vector(assem, data)
+end
+
+function shallow_water_explicit(model, order, Ω, dΩ, dω, qₖ, wₖ, f, g, h₁, u₁, hₘ, uₘ, RTMM, L2MM, dt, leap_frog, τ, P, Q, U, V, R, S)
```
amartinhuertas, line 11: why do you need `qₖ, wₖ` ?

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+  um1          = new_field(V, un)
+  hm2          = new_field(Q, hn)
+  um2          = new_field(V, un)
+  hn, un, ϕ, F = method(model, order, Ω, dΩ, dω, qₖ, wₖ, f, g, hm1, um1, hm2, um2, RTMM, L2MM, dt, false, τ, P, Q, U, V, R, S)
+
+  wn = compute_diagnostics_shallow_water!(model, order, Ω, dΩ, dω, qₖ, wₖ, U, V, R, S, L2MM, H1MM, g, hn, un, ϕ, F, mass, vort, kin, pot, pow, 1, true)
+  
+  # subsequent steps, do leap frog integration (now that we have the state at two previous time levels)
+  for istep in 2:nstep
+    Gridap.FESpaces.get_free_dof_values(hm2) .= Gridap.FESpaces.get_free_dof_values(hm1)
```
amartinhuertas, line 127: `get_free_dof_values` belongs to the public API of `Gridap`. Thus` it is not actually needed to prefix `Gridap.FESpaces.` to each call, i.e., no need to use a fully-qualified name.

src/ShallowWaterIntegrators.jl
amartinhuertas: why do you need `Ω` ?

amartinhuertas: why do you need `order`?

amartinhuertas: can we choose a more precise name for the `shallow_water_explicit` function? For example, one that, apart from referring to the method at hand, tells the reader about what does the function actually does. E.g., does it move the time forward?

test/Williamson2_ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,114 @@
...
+end
+
+l2_err_u = [0.011504453807392859, 0.003188984305055811, 0.0008192298898147198]
+l2_err_h = [0.005636335001937436, 0.0014571807037802682, 0.0003681933640549439]
+
+function forward_step(i, n)
+  order = 1
+  degree = 4
+
+  model = CubedSphereDiscreteModel(n, order+1, radius=rₑ)
```
amartinhuertas, line 52: Remove `order+1` from this call, so that we use the analytical mapping and the AD Jacobian, i.e., we do not approximate the geometry of the sphere.

src/ShallowWaterIntegrators.jl
amartinhuertas: I would try to avoid calling `get_free_dof_values` several times at each loop iteration. 

Instead you can follow the approach in the `GridapGeosciences.jl` implementation of the wave equation SSPRK2 solver. Essentially you extract the reference to the free dof values prior to the loop and use the reference all the way through. 

For example, for `um2` and `um1` you do prior the `for` loop.

`um2v=get_free_dof_values(um2)` 
`um1v=get_free_dof_values(um1)`

Then, at eaach iteration, 

`um2v .= um1v`

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,141 @@
...
+  dt1 = dt
+  if leap_frog
+    dt1 = 2.0*dt
+  end
+
+  # 1.1: the mass flux
+  b₁(v)  = ∫(v⋅u₁*h₁)*dΩ
+  rhs1   = assemble_rhs_vector(U, V, b₁(v))
+  op     = AffineFEOperator(U, V, RTMM, rhs1)
+  F      = solve(op)
```
amartinhuertas, line 35: Ok. This is a critical line from the performance POV. Each time we call `solve` an LU factorizatoin of `RTMM` is computed under the hood. `RTMM` and `L2MM` are not time-dependent, thus this is not needed. 

Do not use `solve`. Instead, compute once `lu` factorization

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L80

and then do forward & backward substitution at each step

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L31

Also note that in that line the `res` vector is re-used over and over again. Your `solve` call also allocates a new vector on each call.

test/Williamson2_ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,114 @@
...
+  reffe_lgn = ReferenceFE(lagrangian, Float64, order+1)
+  S = FESpace(model, reffe_lgn; conformity=:H1)
+  R = TrialFESpace(S)
+
+  # Project the initial conditions onto the trial spaces
+  hc      = CellField(h₀, Ω)
+  a₁(p,q) = ∫(q*p)dΩ
+  b₁(q)   = ∫(q*hc)dΩ
+  op      = AffineFEOperator(a₁, b₁, P, Q)
+  hp      = solve(op)
```
amartinhuertas, line 78: Do we actually need L2 projection here? I expect `interpolate_everywhere`, which solely evaluates FE space Dofs, is going to be much cheaper. See, e.g., 

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L90

for a usage example.

test/Williamson2_ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,114 @@
...
+  U = TrialFESpace(V)
+  reffe_lgn = ReferenceFE(lagrangian, Float64, order)
+  Q = FESpace(model, reffe_lgn; conformity=:L2)
+  P = TrialFESpace(Q)
+  reffe_lgn = ReferenceFE(lagrangian, Float64, order+1)
+  S = FESpace(model, reffe_lgn; conformity=:H1)
+  R = TrialFESpace(S)
+
+  # Project the initial conditions onto the trial spaces
+  hc      = CellField(h₀, Ω)
```
amartinhuertas, line 74: It is not actually necessary to create a `CellField` object in this particular context. You can use  `h0` directly in `b1` below.

In general, `CellField` is necessary when you have an integrand which solely involves operations among analytical functions, e.g., `h0`*`h0`

src/DiagnosticTools.jl
```diff
@@ -42,3 +42,32 @@ function compute_total_mass!(w,L2MM,hh)
...
 end
+
+"""
+  Full diagnostics for the shallow water equations (mass, vorticity, kinetic energy, potential energy, power)
+"""
+function compute_diagnostics_shallow_water!(model, order, Ω, dΩ, dω, qₖ, wₖ, U, V, R, S, L2MM, H1MM, g, h, u, ϕ, F, mass, vort, kin, pot, pow, step, do_print)
+  mass_i = sum(L2MM*Gridap.FESpaces.get_free_dof_values(h))
+  w      = diagnose_vorticity(model, order, Ω, qₖ, wₖ, R, S, U, V, H1MM, u)
+  vort_i = sum(H1MM*Gridap.FESpaces.get_free_dof_values(w))
+  kin_i  = 0.5*sum(∫(h*(u⋅u))dΩ)
```
amartinhuertas, line 53: Dont we have these functions already implemented in `DiagnosticTools.jl` ... if not ... add whatever it is needed there ...

src/DiagnosticTools.jl
```diff
@@ -42,3 +42,32 @@ function compute_total_mass!(w,L2MM,hh)
   mul!(w,L2MM,hh)
   sum(w)
 end
+
+"""
+  Full diagnostics for the shallow water equations (mass, vorticity, kinetic energy, potential energy, power)
+"""
+function compute_diagnostics_shallow_water!(model, order, Ω, dΩ, dω, qₖ, wₖ, U, V, R, S, L2MM, H1MM, g, h, u, ϕ, F, mass, vort, kin, pot, pow, step, do_print)
+  mass_i = sum(L2MM*Gridap.FESpaces.get_free_dof_values(h))
```
amartinhuertas, line 50: Be careful here ... L2MM*Gridap.FESpaces.get_free_dof_values(h) allocates a whole new temporary vector on each call. We want to avoid this. To this end, we can use `mul!` plus input-argument working array, as it is done, e.g., in the following example: 

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L29

I know this is quite low-level conceptually, but the function `mul!` itself belongs to Julia `LinearAlgebra` standard library.

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,121 @@
...
+  # Forward integration of the shallow water equations using a supplied method
+
+  # assemble the mass matrices
+  amm(a,b) = ∫(a⋅b)dΩ
+  H1MM = assemble_matrix(amm, R, S)
+  RTMM = assemble_matrix(amm, U, V)
+  L2MM = assemble_matrix(amm, P, Q)
+
+  # initialise the diagnostics arrays
+  mass = zeros(0)
```
amartinhuertas, line 85: If we know the size of these arrays beforehand, it is better to allocate to this size at the beginning and avoid the frequente usage of `append!`. `append` leads to memory reallocation and copies which are not needed. See

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L99

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L126

for more details

src/ShallowWaterIntegrators.jl
```diff
@@ -0,0 +1,121 @@
...
+  for istep in 2:nstep
+    Gridap.FESpaces.get_free_dof_values(hm2) .= Gridap.FESpaces.get_free_dof_values(hm1)
+    Gridap.FESpaces.get_free_dof_values(um2) .= Gridap.FESpaces.get_free_dof_values(um1)
+    Gridap.FESpaces.get_free_dof_values(hm1) .= Gridap.FESpaces.get_free_dof_values(hn)
+    Gridap.FESpaces.get_free_dof_values(um1) .= Gridap.FESpaces.get_free_dof_values(un)
+
+    hn, un, ϕ, F = method(model, order, Ω, dΩ, dω, qₖ, wₖ, f, g, hm1, um1, hm2, um2, RTMM, L2MM, dt, true, τ, P, Q, U, V, R, S)
+
+    wn = compute_diagnostics_shallow_water!(model, order, Ω, dΩ, dω, qₖ, wₖ, U, V, R, S, L2MM, H1MM, g, hn, un, ϕ, F, mass, vort, kin, pot, pow, istep, true)
+    if mod(istep, dump_freq) == 0
```
amartinhuertas, line 110: For uniformness, I would the same logic to control data generation into files as in 

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L57

test/Williamson2_ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,114 @@
+module Williamson2_ShallowWaterExplicit
```
amartinhuertas, line 1: Add this file to `runtests.jl` ... otherwise we do not test this in Github actions on each push

amartinhuertas (changes requested):
Hi @davelee2804, 

good work! The code is looking prettier, thanks for sticking with me :-) 

I have completed a first round of review. I will complete a second one once you have finished processing my comments. 

@amartinhuertas

davelee2804:
Hi @amartinhuertas , I've tried to address your helpful comments as best I can. The code looks a lot better now thanks! (results are the same as before for the test, I'm kinda surprises that there is no difference in the L2 errors given the solve() has been replaced with the ldiv!() ...)

The only thing I have not addressed (of which I'm aware) is the removal of the `grad_perp_ref_domain()`, and the rest of the routines in the Weak2DCurlOps.jl file. If you feel strongly that these should be removed and replaced with high level physical space terms (and an AD Jacobian) I will do so. My only concern is that without a formal proof that the two are equivalent vorticity will not be conserved and the benefits of compatible FEM will not be gaurunteed.

Let me know what you advise, and thanks again for your review.

amartinhuertas:
Hi @davelee2804,

> Hi @amartinhuertas , I've tried to address your helpful comments as best I can. The code looks a lot better now thanks! (results are the same as before for the test, I'm kinda surprised that there is no difference in the L2 errors given the solve() has been replaced with the ldiv!() ...)

Cool. Thanks for your work! The code looks pretty!  I have marked as resolved all those comments that were already  addressed. 

were the following comments addressed? 

* https://github.com/gridapapps/GridapGeosciences.jl/pull/15#discussion_r700948954, 
* https://github.com/gridapapps/GridapGeosciences.jl/pull/15#discussion_r700965121
* https://github.com/gridapapps/GridapGeosciences.jl/pull/15#discussion_r700973391

In general, in PRs, if we decide not to go ahead with a recommendation, we also typically write some words on why not on the corresponing comment text box.  The code reviewer might not have the whole picture of the work done, and may provide a missleading recommendation due to that. Like the rebuttal letter to the paper referees :-) 

> The only thing I have not addressed (of which I'm aware) is the removal of the grad_perp_ref_domain(), and the rest of the routines in the Weak2DCurlOps.jl file. If you feel strongly that these should be removed and replaced with high level physical space terms (and an AD Jacobian) I will do so. My only concern is that without a formal proof that the two are equivalent vorticity will not be conserved and the benefits of compatible FEM will not be gaurunteed.

First of all, just in case it is not clear, we can either use isoparametric or analytical mapping independently of what we want to do with the right hand side of the weak div perp operator.  Second,  I would go for the high level API term. Numerically, they seem to be equivalent (up to a minus sign). There is no mathematical proof yet they are equivalent, that is right.

At this point, I you are still concerned, what I suggest you to do is the following. 

1. Let us create a branch of the current branch with the current implementation, `shallow_water_explicit_grad_perp_ref_vorticity`. 
2. Let us transform the current implementation in `shallow_water_explicit` using the high level API terms. You will see that the code will look even prettier, more general, and abstract.
3. Let us compare systematically with a benchmark problem that you are particular worried about the results of the codes in branches 1 and 2. This would provide more confidence on their equivalence. 
4.  If it turns out that there are differences among 2. and 3., let us think how can we abstract the `grad_perp_ref_domain` term using the high level API. Hopefully they turn to be equivalent and we just have to merge 2. into main.  

What do you think?

test/Williamson2_ShallowWaterExplicit.jl
davelee2804: Currently still using the isoparametric Jacobian in order to reduce under integration of the metric terms. However I take your point that there will still be under-integration of the Gammian even for the isoparametric Jacobian, so I will change over to the AD Jacobian...

src/ShallowWaterIntegrators.jl
davelee2804: I thought I had addressed this one? Sorry, I might need some more clarification here...

davelee2804:
Hey @amartinhuertas , sounds like a good plan. I will make this secondary branch for the grad_perp_ref_domain, and then revert the existing branch to the high level physical space version as you suggest. Since we are computing vorticity conservation as a diagnostic we can monitor this for any bias with the physical space implementation and revisit if necessary.

Sorry for not commenting on the changes I had not made, I'll be more careful with this in future....

davelee2804:
Hi @amartinhuertas , I've made the grad_perp branch as you suggested, then removed the low level reference operators from the shallow_water_explicit branch and replaced these with the high level physical space terms as you suggested. The errors are similar but not the same. Here are the conservation and l2 errors for the final 10 time steps for both configurations using the C16 mesh and RT1 elements, first the high level physical space curl:

```
step | mass conservation | vorticity conservation | kinetic energy | potential energy | energy conservation | power
70	-1.9114360408403353e-15	5.960464477539063e-8	6.633793145418441e20	1.477262020091013e22	-6.469974972230993e-10	2.9468087973284225e9
71	-1.9114360408403353e-15	1.7881393432617188e-7	6.633793153306743e20	1.4772620200009785e22	-6.542217757846671e-10	5.013208853950043e9
72	-1.9114360408403353e-15	0.0	6.633793164808813e20	1.4772620198748801e22	-6.613983670972954e-10	7.013473197860233e9
73	-1.9114360408403353e-15	-1.7881393432617188e-7	6.63379317773161e20	1.4772620197346446e22	-6.685294449387136e-10	7.772140080217796e9
74	-1.9114360408403353e-15	0.0	6.633793188754459e20	1.4772620196134932e22	-6.756057707535717e-10	6.685436270421591e9
75	-1.9114360408403353e-15	5.960464477539063e-8	6.633793195407092e20	1.4772620195361106e22	-6.82638756874949e-10	4.243648649419777e9
76	-1.9114360408403353e-15	0.0	6.633793197822413e20	1.4772620195011748e22	-6.896241916084949e-10	1.8942620096129227e9
77	-1.9114360408403353e-15	-3.5762786865234375e-7	6.633793199121693e20	1.47726201947747e22	-6.965637052875065e-10	1.280609604765152e9
78	-2.1238178231559282e-15	-5.960464477539063e-8	6.633793204058822e20	1.4772620194174609e22	-7.034553958564704e-10	3.2850473646575394e9
79	-1.9114360408403353e-15	5.364418029785156e-7	6.633793216577788e20	1.477262019281706e22	-7.102999426209271e-10	7.441219225807159e9
80	-2.1238178231559282e-15	1.1920928955078125e-7	6.633793237602512e20	1.4772620190609624e22	-7.170997910808223e-10	1.2075524820265934e10
n=16,	err_u: 0.0007415233680146857,	err_h: 0.00036813020391681306
```
...and here are the low level reference space grad-perp conservation and l2 errors:
```
step | mass conservation | vorticity conservation | kinetic energy | potential energy | energy conservation | power
70	-1.9114360408403353e-15	0.0	6.633793145418448e20	1.4772620200910134e22	-6.46997089639775e-10	2.9468087698963394e9
71	-1.9114360408403353e-15	1.7881393432617188e-7	6.633793153306746e20	1.4772620200009787e22	-6.542216399235591e-10	5.013208836234894e9
72	-1.9114360408403353e-15	1.1920928955078125e-7	6.633793164808811e20	1.4772620198748807e22	-6.613978236528631e-10	7.0134731802047e9
73	-1.9114360408403353e-15	1.1920928955078125e-7	6.633793177731617e20	1.4772620197346446e22	-6.685294449387136e-10	7.772140071719622e9
74	-1.9114360408403353e-15	1.7881393432617188e-7	6.633793188754468e20	1.4772620196134934e22	-6.756053631702474e-10	6.685436286510313e9
75	-1.9114360408403353e-15	2.384185791015625e-7	6.633793195407095e20	1.4772620195361104e22	-6.826388927360571e-10	4.2436486332578735e9
76	-2.1238178231559282e-15	2.980232238769531e-7	6.633793197822415e20	1.4772620195011752e22	-6.896237840251706e-10	1.894261969870346e9
77	-1.9114360408403353e-15	2.384185791015625e-7	6.633793199121695e20	1.4772620194774694e22	-6.965641128708307e-10	1.2806096133585129e9
78	-2.336199605471521e-15	0.0	6.633793204058827e20	1.477262019417461e22	-7.034552599953623e-10	3.2850473504682846e9
79	-1.9114360408403353e-15	2.384185791015625e-7	6.633793216577798e20	1.477262019281705e22	-7.103002143431434e-10	7.441219187366409e9
80	-2.1238178231559282e-15	1.7881393432617188e-7	6.63379323760251e20	1.4772620190609633e22	-7.1709924763639e-10	1.2075524782441132e10
n=16,	err_u: 0.0007415233680147283,	err_h: 0.0003681302039168138
```
Since they there is not obvious bias in the vorticity conservation errors, and they are the same magnitude in both cases, we are _probably_ OK, though I will continue to monitor these vorticity conservation errors for more advanced tests...

Are there any more outstanding issues I should address?

Thanks again.

amartinhuertas:
Hi @davelee2804, 

nice work! 

If I understand correctly the results in the tables (and as you say), it seems both quantitatively and qualitatively that both approaches achieve the desired outcome right? It is surprising that the vorticity becomes exactly 0.0 in some steps (happens in both cases).

If you still feel that a single experiment is still not enough, we may want to systematize experiments with the help of DrWatson.jl as well. I expect this work to be also necessary for the future, to compare systematically different algorithms and parameter values (mesh resolution, time step, FE order, geometrical order, etc.). I could help with this.

In regards to the PR, I will start now a new round of revision, with additional comments.

Thanks for your work!

src/DiagnosticTools.jl
amartinhuertas: Have you addressed this comment?

src/Operators.jl
```diff
@@ -84,3 +84,12 @@ function gradient_unit_sphere(v)
      Jθϕ⋅inv(Gθϕ)⋅(∇(v))(θϕ)
   end
 end
+
+"""
+Reference domain 2D ∇⟂
```
amartinhuertas, line 89: I guess this function is not needed in this branch, right?

src/GridapGeosciences.jl
```diff
@@ -22,4 +24,9 @@ module GridapGeosciences
   export cartesian_to_spherical_matrix
   export Eₖ, Eₚ, Eₜ
   export compute_kin_to_pot!, compute_pot_to_kin!, compute_total_mass!
+  export compute_diagnostics_shallow_water!
+  export grad_perp_ref_domain
```
amartinhuertas, line 28: Remove this export

src/DiagnosticTools.jl
```diff
@@ -42,3 +42,37 @@ function compute_total_mass!(w,L2MM,hh)
...
+  kin[step]  = kin_i
+  pot[step]  = pot_i
+  pow[step]  = pow_i
+
+  if do_print
+    # normalised conservation errors
+    mass_norm = (mass_i-mass[1])/mass[1]
+    vort_norm = vort_i-vort[1]
+    en_norm   = (kin_i+pot_i-kin[1]-pot[1])/(kin[1]+pot[1])
+    println(step, "\t", mass_norm, "\t", vort_norm, "\t", kin_i, "\t", pot_i, "\t", en_norm, "\t", pow_i)
```
amartinhuertas, line 73: FYI ... for the pretty printing of tabular data, there is this very nice package

https://ronisbr.github.io/PrettyTables.jl/stable/

The only caveat is that I think the package does not allow you to plot on a row by row basis progressively but in one shot :-(

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  wn     = FEFunction(S, copy(get_free_dof_values(f)))
+  # first step, no leap frog integration
+  shallow_water_explicit_time_step!(model, dΩ, dω, f, g, hm1, um1, hm2, um2, hp, up, RTMMchol, L2MMchol, dt, false, τ, Q, V, R, S, hn, un, ϕ, F)
+  if mod(1, diag_freq) == 0
+    compute_diagnostics_shallow_water!(model, dΩ, dω, S, L2MM, H1MM, H1MMchol, h_tmp, w_tmp, g, hn, un, ϕ, F, mass, vort, kin, pot, pow, 1, true, wn)
+  end
+  
+  # subsequent steps, do leap frog integration (now that we have the state at two previous time levels)
+  for istep in 2:nstep
+    get_free_dof_values(hm2) .= get_free_dof_values(hm1)
```
amartinhuertas, line 156: See this comment: https://github.com/gridapapps/GridapGeosciences.jl/pull/15#discussion_r700948954 

Is there any reason why you did not address it?

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  copy!(get_free_dof_values(h₂), rhs8)
+  ldiv!(L2MMchol, get_free_dof_values(h₂))
+end
+
+function shallow_water_time_stepper(model, order, degree, h₀, u₀, f₀, g, nstep, diag_freq, dump_freq, dt, τ)
+  # Forward integration of the shallow water equations using a supplied method
+  Ω = Triangulation(model)
+  dΩ = Measure(Ω, degree)
+  dω = Measure(Ω, degree, ReferenceDomain())
+  quad_cell_point = get_cell_points(dΩ.quad)
```
amartinhuertas, line 87: I would say these four lines are no longer needed right ?

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  ldiv!(RTMMchol, get_free_dof_values(F))
+  # 1.2: the bernoulli function
+  b₂(q)  = ∫(q*(0.5*u₁⋅u₁ + g*h₁))*dΩ
+  rhs2   = assemble_vector(b₂, Q)
+  copy!(get_free_dof_values(ϕ), rhs2)
+  ldiv!(L2MMchol, get_free_dof_values(ϕ))
+  # 1.3: the potential vorticity
+  a₁(r,s) = ∫(s*h₁*r)dΩ
+  c₁(s)   = ∫(perp(n,∇(s))⋅(u₁) + s*f)dΩ
+  H1h     = assemble_matrix(a₁, R, S)
```
amartinhuertas, line 38: Although correct, this is not the most efficient way to assemble this problem. If you separate the assembly of the matrix and the right hand side, you will need two different loops over the mesh (once for the matrix and another for the vector). Better to fuse both loops in a single one. To this end,

(1) Remove `assemble_matrix` and `assemble_vector` calls.
(2) Replace `AffineFEOperator(R, S, H1h, rhs_q₁)` by `AffineFEOperator(a1, b1, R, S)`

Does this make sense?

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  ldiv!(RTMMchol, get_free_dof_values(F))
+  # 2.2: the bernoulli function
+  b₆(q)  = ∫(q*((u₁⋅u₁ + u₁⋅uₚ + uₚ⋅uₚ)/6.0 + 0.5*g*(h₁ + hₚ)))*dΩ
+  rhs6   = assemble_vector(b₆, Q)
+  copy!(get_free_dof_values(ϕ), rhs6)
+  ldiv!(L2MMchol, get_free_dof_values(ϕ))
+  # 2.3: the potential vorticity
+  a₂(r,s) = ∫(s*hₚ*r)dΩ
+  c₂(s)   = ∫(perp(n,∇(s))⋅(uₚ) + s*f)dΩ
+  H2h     = assemble_matrix(a₂, R, S)
```
amartinhuertas, line 66: see https://github.com/gridapapps/GridapGeosciences.jl/pull/15/files#r702352434

src/GridapGeosciences.jl
```diff
@@ -8,9 +8,11 @@ module GridapGeosciences
   include("Operators.jl")
   include("CoordinateTransformations.jl")
   include("DiagnosticTools.jl")
+  include("ShallowWaterExplicit.jl")
   export rₑ, Ωₑ, g, f
   export CubedSphereDiscreteModel
   export perp,⟂
+  export grad_perp
```
amartinhuertas, line 15: Remove this export

amartinhuertas:
Hi @davelee2804 ... I am going out for a walk. I will continue later adding more comments to my revision.

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  ldiv!(RTMMchol, get_free_dof_values(u₂))
+  # 2.5: solve for the final depth
+  b₈(q)  = ∫(q*h₁)dΩ - ∫(dt*q*DIV(F))*dω
+  rhs8   = assemble_vector(b₈, Q)
+  copy!(get_free_dof_values(h₂), rhs8)
+  ldiv!(L2MMchol, get_free_dof_values(h₂))
+end
+
+function shallow_water_time_stepper(model, order, degree, h₀, u₀, f₀, g, nstep, diag_freq, dump_freq, dt, τ)
+  # Forward integration of the shallow water equations using a supplied method
```
amartinhuertas, line 83: Back in business ... Is this comment obsolete? What is the "supplied method" ?

test/runtests.jl
```diff
@@ -7,5 +7,6 @@ using Test
   @time @testset "WeakDivPerpTests" begin include("WeakDivPerpTests.jl") end
   @time @testset "LaplaceBeltramiCubedSphereTests" begin include("LaplaceBeltramiCubedSphereTests.jl") end
   @time @testset "WaveEquationCubedSphereTests" begin include("WaveEquationCubedSphereTests.jl") end
+  @time @testset "Williamson2_ShallowWaterExplicit" begin include("Williamson2_ShallowWaterExplicit.jl") end
```
amartinhuertas, line 10: Remove underscore from test set and file name. Add the postfix `Tests` at the end to both names.

amartinhuertas:
Comment out temporarily `Manifest.toml` from `.gitignore` and commit and push `Manifest.toml`. Note that we are temporarily pointing to a particular branch in `Gridap.jl` github repo. (while these devs are not released, currently under revision). Thus, this is needed to let the tests pass in Github CI actions. By the way, I would stop using `[ci skip]` for a while in this PR so that we double check that the tests pass in Github CI actions. We want to merge if and only if the tests are passing there.

src/ShallowWaterIntegrators.jl
amartinhuertas: > I thought I had addressed this one? Sorry, I might need some more clarification here...

Use the same optional keyword parameters. i.e, `write_results`, `out_dir`, `out_period`, while adjusting the appropriate default values for the function at hand. See for more details: https://github.com/gridapapps/GridapGeosciences.jl/blob/master/test/WaveEquationCubedSphereTests.jl#L59
Look for these three variables there and replicate the logic in your function. 

Note that: (1) output into data files can be activated/deactivated on demand. E.g., for the tests in Github actions we do not want to generate data files, no one is going to visualize them. (2) one may select the name of the directory where to output the data files, although a meaninful default one is generated if not provided (3) the directory where the output data is placed is created automatically (and truncated/removed if it already exists). (4) We use writevtk.jl to generate a PVD file (like some sort of XML header file which points to all time steps data). This latter file is required by Paraview in order to apply some time dependent filters, and can be seen as a sort of  file that glues together the transient simulation.

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
...
+  # h₁         : fluid depth at current time level
+  # u₁         : fluid velocity at current time level
+  # hₘ         : fluid depth at previous time level (for leap-frogging the first step)
+  # uₘ         : fluid velocity at previous time level (for leap-frogging the first step)
+  # RTMM       : H(div) mass matrix, ∫β⋅βdΩ, ∀β∈ H(div,Ω)
+  # L2MM       : L² mass matrix, ∫γγdΩ, ∀γ∈ L²(Ω)
+  # dt         : time step
+  # leap_frog  : do leap frog time integration for the first step (boolean)
+  # dΩ         : measure of the elements
+
```
amartinhuertas, line 17: I am still concerned about the performance of this function, i.e., `shallow_water_explicit_time_step!`. I think it can still be significanlty improved. In particular, as the mesh does not change along the simulation (e.g., we do not currently have dynamic mesh adaptation),  the size of matrices and vectors, and the symbolic sparsity pattern of matrices remains the same.  Thus, there is a lot of potential for memory to be reused among calls to this function. Currently we are dynamically allocating at each call:

1. Eight vectors, namely `rhs1`-`rhs8`.
2. The two coefficient matrices of the two potential vorticity linear systems. (steps 1.3 and 2.3).
3. The two LU factorizations  of the coefficient matrices in 2. Note that, although we need to compute these LU factorizations at each step (as the entries of the matrices in 2. change), we may try to reuse the symbolic sparsity pattern of the L and U factors, and the memory buffers of the L and U factors. 

At present, I know how to address 1. and 2. I think that 3. can also be addressed using low level Gridap functions. Let us try to first address 1. and 2. in a different comment box.

amartinhuertas: In order to address 1. and 2. we can use in-place `assemble_vector!` and `assemble_matrix_and_vector!` assembly functions.  

The idea is that we pass work array vectors (do we actually need eight `rhs` vectors?) and work matrices (do we actually need to matrices or just a single one?)  to `shallow_water_explicit_time_step!` and we call these functions to perform assembly in-place instead of generating whole new objects each time. Note that generating a sparse matrix from scratch is much more expensive than reusing an already generated matrix, as with the former we need to compute also the sparsity pattern, i.e., where the nonzero entries are located. 

It turns out that the functions that are more appropiate in your particular context are not already available in the latest version of Gridap. However, I have developed them in the following PR:  https://github.com/gridap/Gridap.jl/pull/652, which is currently under review and will be merged soon. See, e.g., 

https://github.com/gridap/Gridap.jl/blob/developed_additional_highlevel_assembly_functions/test/FESpacesTests/AssemblersTests.jl#L78

for examples of usage of these functions.

Finally, in order to help you out, I have pulled all these changes into the Gridap.jl branch your are pointing to from `GridapGeosiences.jl`. Thus, if you just run `up Gridap` from the package manager, you will have these two functions are your disposal.

amartinhuertas: In order to address 3., you can look at lines 246-247 of the following source file of `Gridap.jl` 

https://github.com/gridap/Gridap.jl/blob/master/src/Algebra/LinearSolvers.jl#L249

In particular, we are interested into calling the `numerical_setup!` and `solve!` functions inside the function. 

A short snippet is as follows (where `A` is assumed to be an assembled sparse matrix):

```julia
ls = LUSolver()
lss = symbolic_setup(ls,A)
lsn = numerical_setup(lss,A) 
solve!(x,lsn,b)                         # Solve A x = b 
                                                   # Change the entries of `A` by whatever process ...
numerical_setup!(lsn,A)     #  Reucse L,U memory buffers to recompute LU factorization in-place
solve!(x,lsn,b)                         # Solve Ax = b with the updated A
```

Let me know if you face issues with this ...

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,171 @@
+function shallow_water_explicit_time_step!(model, dΩ, dω, f, g, h₁, u₁, hₘ, uₘ, hₚ, uₚ, RTMMchol, L2MMchol, dt, leap_frog, τ, Q, V, R, S, h₂, u₂, ϕ, F)
```
amartinhuertas, line 1: Let us finish this round of optimization first, but I would like that we give some more profound thinking towards homogeneization and satisfaction of coding guidelines to the signatures of the different functions that we write in a third round. Things that come into my mind:

- Those arguments which are modified by side-effecting functions (i.e., marked by `!`  by convention) are usually placed first. 
- Can we reduce the number of arguments? E.g., `g` is a constant defined in `GridapGeosciences.jl`, I think there is no need to pass it. 
- All finite element functions corresponding to the different time steps can be passed as a tuple. 
-  Why do we need `R`, and not `P` and `U`. All FE spaces can be passed also as a tuple. 
- All cholesky factorizations/solvers needed can be also passed as a tuple.
- What else?

amartinhuertas (changes requested):
Hi @davelee2804 ... I am finished with my second round of review.  Do not hesitate to contact me is tthere are any issues while addressing them. We are closer!

src/ShallowWaterExplicit.jl
davelee2804: When I tried to address this one it did not work (ie: the values did not change between time steps)

src/ShallowWaterExplicit.jl
davelee2804: I need the top three to pass down to the explicit solver, but not the others below, have removed...

src/ShallowWaterExplicit.jl
davelee2804: Yes, I didn't know the difference between these two. Will do, thanks!

src/ShallowWaterExplicit.jl
davelee2804: Great spotting!

src/ShallowWaterExplicit.jl
davelee2804: I have not removed `g` as there may be scenarios where we want to override the default value (three that come to mind are non-dimensionalised configurations, exoplanets (with a different `g`) and deep atmosphere approximations (where `g` can vary with radius). Hope that's OK

Also I haven't packaged everything up into tuples as not all the FESpaces or Cholesky factorisations are needed by the time stepper, and I'm wondering if there a performance cost to passing unneeded ones in. That said if you think it is better I'm happy to make a tuple out of the ones that are needed and pass those in....

src/ShallowWaterExplicit.jl
amartinhuertas: That's interesting ... I do not understand how this cannot actually work. The only reason it cannot work is that the pointer that the FE functions hm2, hm1, etc. hold inside changes along the process, but I do not understand why is this the case in your code. 

Can you please add the following two lines at the beginning of the for loop:

```
println(objectid(get_free_dof_values(hm2)))
println(objectid(get_free_dof_values(hm1)))
```

Does the indentifier changes along the process? 

In any case, looking more carefully at this code, note that you are copying the entries. Is that really necessary? Can we instead use pointer swapping? (using the `FEFunction` constructor wisely)

src/ShallowWaterExplicit.jl
amartinhuertas: > I have not removed g as there may be scenarios where we want to override the default value (three that come to mind are non-dimensionalised configurations, exoplanets (with a different g) and deep atmosphere approximations (where g can vary with radius). Hope that's OK

Ok, that makes sense. 

> Also I haven't packaged everything up into tuples as not all the FESpaces or Cholesky factorisations are needed by the time stepper, and I'm wondering if there a performance cost to passing unneeded ones in. That said if you think it is better I'm happy to make a tuple out of the ones that are needed and pass those in....

It's ok, I am concerned about the very large number of arguments. But the tuple thing does not actually solve the problem. (it is just a dirty patch).  The function is may be doing too many things. Just an idea ... if you do not see it, forget it ... may we split this function into more functions, i.e., one per each of the steps, so that we end up with simpler, reusable functions in other scenarios? Note that large functions are hard to reuse.

src/ShallowWaterExplicit.jl
amartinhuertas: Ok I see how you have solved this comment! Good approach!

amartinhuertas:
@davelee2804 ... FYI ... tests are failing in this branch at Github Actions. I am positive it is related to this:

https://github.com/gridapapps/GridapGeosciences.jl/pull/15#issuecomment-913085099

Also take into account the modified signatures for `assemble_vector!` and `assemble_matrix_and_vector!` signatures.

davelee2804:
Hey @amartinhuertas , hopefully I've addressed all the issues you raised (except for the ones I have already commented on). The CI test is failing at:
https://github.com/gridapapps/GridapGeosciences.jl/blob/shallow_water_explicit/src/ShallowWaterExplicit.jl#L27
but presumably this is because the CI testing is referencing the older version of `assemble_vector!()` with the different argument ordering.
Please let me know if I've missed anything, and thanks again for the super helpful and educational review!

src/ShallowWaterExplicit.jl
```diff
@@ -0,0 +1,155 @@
...
+  # first step, no leap frog integration
+  shallow_water_explicit_time_step!(hn, un, ϕ, F, q1, q2, model, dΩ, dω, f, g, hm1, um1, hm2, um2, hp, up, RTMMchol, L2MMchol, H1h, H1hchol, dt, false, τ, Q, V, R, S)
+  initialize_csv(joinpath(out_dir,"swe_diagnostics.csv"), "time", "mass", "vorticity", "kinetic", "potential", "power")
+  if mod(1, diag_freq) == 0
+    compute_diagnostics_shallow_water!(wn, model, dΩ, dω, S, L2MM, H1MM, H1MMchol, h_tmp, w_tmp, g, hn, un, ϕ, F, 1, dt, true, out_dir)
+  end
+  
+  # subsequent steps, do leap frog integration (now that we have the state at two previous time levels)
+  for istep in 2:nstep
+    hm2,hm1,hn = hm1,hn,hm2
```
amartinhuertas, line 142: Ok, this is definitely the way to go. Pointer swapping, no copies ...

amartinhuertas:
>Hey @amartinhuertas , hopefully I've addressed all the issues you raised (except for the ones I have already commented on). The CI test is failing at:
https://github.com/gridapapps/GridapGeosciences.jl/blob/shallow_water_explicit/src/ShallowWaterExplicit.jl#L27
but presumably this is because the CI testing is referencing the older version of assemble_vector!() with the different argument ordering.
Please let me know if I've missed anything, and thanks again for the super helpful and educational review! 

I will fix the failing test, no worries.

On the other hand, have you addressed this?

https://github.com/gridapapps/GridapGeosciences.jl/pull/15#discussion_r700965121

Finally, I would give a shot to splitting your long functions into subfunctions. I think having functions that only do one thing is essential. If you do not find sufficiently confident how to do it, I can do it. Let me know.

src/DiagnosticTools.jl
davelee2804: Thanks for spotting. I have generalised the `E_k` function so that this can take h as a FEFunction (so that it can be used to compute KE for the shallow water equations)

davelee2804:
Hey @amartinhuertas , I don't know how to split up the explicit function. I know it is long, and as a two stage function is make sense to split it. However each of the two stages is different and relies on a different about of data (the first stage relies on a single time level of data, the second stage on two).

If you have any idea on how to split it I'm happy to try, but there is no obvious way to me (It is not like this is straight Runge-Kutta where you do the same thing on different data at each stage)

src/ShallowWaterIntegrators.jl
amartinhuertas: I have completed this task in this commit https://github.com/gridapapps/GridapGeosciences.jl/pull/15/commits/2ab0486128709dbec189c005b0f29cfb10f4f147

There were still pending devs :-)

amartinhuertas:
> Hey @amartinhuertas , I don't know how to split up the explicit function. I know it is long, and as a two stage function is make sense to split it. However each of the two stages is different and relies on a different amount of data (the first stage relies on a single time level of data, the second stage on two).
> If you have any idea on how to split it I'm happy to try, but there is no obvious way to me (It is not like this is straight Runge-Kutta where you do the same thing on different data at each stage)

You can see the result of the work here: https://github.com/gridapapps/GridapGeosciences.jl/blob/60313734c8d369d57fa2803681e88cff492ef83d/src/ShallowWaterExplicit.jl#L68

Let me know if that makes sense. 

Please note that the function has been split into many different reusable functions which only do one thing.  I guess that these functions might be used for other solvers as well. Also note that it was impossible to reuse code from the previous function without replicating it.

amartinhuertas (approved):
@davelee2804 ... nice work ... I am already happy with the current code. Feel free to merge once you have taken a look at the latest changes.

davelee2804:
Thanks for all your help here @amartinhuertas , I've learnt a lot! I made one final change before merging, the L2 errors were different in the last couple of decimals (perhaps the order of operations changed on something). Nothing big enough to trigger a test failure (set at 1.0e-12), and perhaps we are not bit-4-bit on different machines anyway, but just for consistency.

davelee2804:
Hey @amartinhuertas , the merge just failed. Is this related to the commit you just made?

amartinhuertas:
> Hey @amartinhuertas , the merge just failed. Is this related to the commit you just made?

I do not see that it failed (still on the way). Can you send the link to Github Actions of the failed job?

amartinhuertas:
> I made one final change before merging, the L2 errors were different in the last couple of decimals (perhaps the order of operations changed on something). Nothing big enough to trigger a test failure (set at 1.0e-12), and perhaps we are not bit-4-bit on different machines anyway, but just for consistency.

FYI ... Julia offers the built-in `isapprox` function to facilitate the implementation of this kind of floating point equality tests.

davelee2804:
Hmm, I can't see it in the actions, and the error message is no longer here, I'm confused. I'll wait til your CI tests have run and try again...

davelee2804:
Merged now :)

amartinhuertas:
Note that the order of the arguments has changed for some functions ... When you pull from master into the other brnaches this will affect you.

davelee2804:
OK, thanks I'll try and keep that in mind

PR #17, opened 2021-09-07
Wave eq csv output

cbladwell:
Hi @amartinhuertas @davelee2804 I've just opened this pull request to discuss saving output data to `.csv`. This can either be merged and used in the NSWE code or we can use this as the starting point for work on the diagnostics


davelee2804:
Hey @cbladwell , thanks for doing this. Your approach looks good to me. I will emulate this in the shallow_water_explicit branch as well

cbladwell:
No worries. Should we move these functions into `DiagnosticTools.jl` and you merge into your shallow water branch?

davelee2804:
Yes, I think that is a good Idea. Will add these to the diagnostics tonight

Project.toml
```diff
@@ -4,13 +4,15 @@ authors = ["Santiago Badia <santiago.badia@monash.edu>", "Christopher Bladwell <
 version = "0.1.0"
 
 [deps]
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 Gridap = "56d4f2e9-7ea1-5844-9cf6-b9c51ca7ce8e"
 JLD = "4138dd39-2aa7-5051-a626-17a0bb65d9c8"
```
amartinhuertas, line 10: Remove JLD as a dependency

test/WaveEquationCubedSphereTests.jl
```diff
@@ -7,6 +7,8 @@ using Plots
 using LinearAlgebra
 using WriteVTK
 using JLD
```
amartinhuertas, line 9: Remove using JLD

test/WaveEquationCubedSphereTests.jl
```diff
@@ -139,11 +154,22 @@ function solve_wave_equation_ssrk2(
       vtk_save(pvd)
       generate_energy_plots(out_dir,N,ke,pe,kin_to_pot,pot_to_kin)
       # save global scalar snapshots
-      save(joinpath(out_dir,"wave_eq_geosciences_data.jld"), "hn_dot_div_un", kin_to_pot,
-                                                "un_dot_grad_hn", pot_to_kin,
-                                                "mass", mass,
-                                                "kinetic", ke,
-                                                "potential", pe)
+      # save(joinpath(out_dir,"wave_eq_geosciences_data.jld"), "hn_dot_div_un", kin_to_pot,
```
amartinhuertas, line 157: Remove old code ...

amartinhuertas:
Nice work! @cbladwell ... just a couple of additional comments 

1. Remove png and data files from the repo.
2. Tests are failing, why? Solve it.
3. I think it is better to leave write_results=false by default so that we do not generate the data in Github actios CI (I am not sure if this can cause trouble).

davelee2804:
Hey @cbladwell , I have one comment. It looks like you are writing the csv file at the end of the time step iteration. I think it is better if the file is instead appended each time step. The reason is that if the code crashes (or we hit wall time on the HPC) before the end of the time step loop we will not have any diagnostics data to inspect. Could you move the code to append the file at each time step? Does that sound ok?

amartinhuertas:
> Hey @cbladwell , I have one comment. It looks like you are writing the csv file at the end of the time step iteration. I think it is better if the file is instead appended each time step. The reason is that if the code crashes (or we hit wall time on the HPC) before the end of the time step loop we will not have any diagnostics data to inspect. Could you move the code to append the file at each time step? Does that sound ok?

hi @davelee2804 ... I agree this would be desirable, however, I am not sure if this is possible with CSV.jl. Perhaps @cbladwell can tell better.

cbladwell:
>  It looks like you are writing the csv file at the end of the time step iteration. I think it is better if the file is instead appended each time step. The reason is that if the code crashes (or we hit wall time on the HPC) before the end of the time step loop we will not have any diagnostics data to inspect. Could you move the code to append the file at each time step? Does that sound ok?

Yes I agree, it would be computationally better to write at each output interval (as with the `.vtu`s) rather than storing the fields in memory. I think it is possible to do this with CSV.jl

test/WaveEquationCubedSphereTests.jl
cbladwell: Updated. I've removed the `TypedTables`  for `DataFrames` as well

cbladwell:
>  Tests are failing, why? Solve it.

I think this error is in the Williamson tests. I can investigate the error further but I think it is fixed in the most recent Williamson test code

amartinhuertas:
Thanks for your work @cbladwell ! 

> I think this error is in the Williamson tests. I can investigate the error further but I think it is fixed in the most recent Williamson test code

Notet that now (latest commit) you have a different error: https://github.com/gridapapps/GridapGeosciences.jl/runs/3532983666#step:6:224

The error that you had before was different: 

https://github.com/gridapapps/GridapGeosciences.jl/runs/3529272826?check_suite_focus=true#step:6:724

I think this latter one is caused because write_results=true, which in turns triggers the plot command that generates an error because there is no X terminal to connect to. 

I am aware there are errors in other tests, but these are puzzling me ... it is like everthing failed!

amartinhuertas:
> I am aware there are errors in other tests, but these are puzzling me ... it is like everthing failed!

Ok, I know what it is failing. Your Manifest.toml file is wrong:

https://github.com/gridapapps/GridapGeosciences.jl/blob/wave_eq_csv_output/Manifest.toml#L325

it should point to a particular branch of Gridap.jl. Look how this file looks like at `master`

https://github.com/gridapapps/GridapGeosciences.jl/blob/master/Manifest.toml#L312

amartinhuertas:
See also changes to manifold.toml in this branch 

https://github.com/gridapapps/GridapGeosciences.jl/pull/17/files#diff-6d1205e03b7a43c01ed2ac8b8c83eb58c6f3a3b1360769e18644245abee8c565R325

cbladwell:
> it should point to a particular branch of Gridap.jl. Look how this file looks like at `master`

Thanks! There were a couple of errors in the Manifest I have updated now. 

@davelee2804  sorry for the delay, you should be able to write to csv now

PR #18, opened 2021-09-10
Rosenbrock solver for the SWEs

amartinhuertas:
Creating PR so that we can start discussing this work as needed ... WIP ...


davelee2804:
Here are the outstanding tasks that I'm aware of (there may be others)
- [ ] Assemble the approximate Jacobian for the `[u,h]` coupled system with (`B`) and without (`A`) the mass matrices
- [ ] Solve for `du, dh` in `shallow_water_rosenbrock_time_step!`
- [ ] Refactor the velocity and depth field residual assembly into separate functions 
- [ ] Convergence test using the Williamson2 configuration (as done for the explicit solver), can probably use a time step 10x that of the explicit solver
- [ ] Run with the Galewsky test case for 20 days. Should be stable for a time step of 360s, probably a bit larger

davelee2804:
Hey @cbladwell , you've done some nice work here, looks like we're almost finished, well done! I'm just setting up to run your code now, but here are a few (minor) comments in the interim:
- looks like this has been defined in a couple of places, might want to consolidate... https://github.com/gridapapps/GridapGeosciences.jl/blob/shallow_water_rosenbrock/src/ShallowWaterRosenbrock.jl#L10
- `Amat*get_free_dof_values(duh1)` may create a new vector (not sure), much be safer to create a work vector at the beginning of the simulation, and pass this and then do as `mul(my_work_vec, Amat, get_free_dof_values(duh1))`: https://github.com/gridapapps/GridapGeosciences.jl/blob/shallow_water_rosenbrock/src/ShallowWaterRosenbrock.jl#L84

cbladwell:
I have update the multiplication by `Amat.

> looks like this has been defined in a couple of places, might want to consolidate... https://github.com/gridapapps/GridapGeosciences.jl/blob/shallow_water_rosenbrock/src/ShallowWaterRosenbrock.jl#L10

Agreed. I haven't moved it yet but this could be an internal function `_clone_fe_function` in `Operators.jl`

davelee2804:
Thanks @cbladwell , I just fixed up a bug and made a minor optimisation and pushed the changes...

davelee2804:
Galewsky test case results look good (instability & conservation errors). Ready for revie
vort_day6
w
vort_cons

mass_cons
ener_cons

driver/GalewskyShallowWaterExplicit.jl
```diff
@@ -0,0 +1,79 @@
...
+using Gridap
+using GridapGeosciences
+
+# Solves the Galewsky test case for the shallow water equations on a sphere
+# of physical radius 6371220m. Involves a shear flow instability of a zonal 
+# jet triggered by an initial gravity wave.
+# reference:
+#   Galewsky, Scott and Polvani (2004) Tellus, 56A 429-440
+
+function uθ(θϕr)
```
amartinhuertas, line 12: I would move these three function definition statements to its own julia source file, e.g., `GalewskyInitialConditions.jl`, located at the `driver` folder. Then, we can `include(GalewskyInitialConditions.jl)` from both driver files and avoid code replication.

test/Williamson2ShallowWaterRosenbrockTests.jl
```diff
@@ -0,0 +1,86 @@
...
+
+# Solves the steady state Williamson2 test case for the shallow water equations on a sphere
+# of physical radius 6371220m. Involves a modified coriolis term that exactly balances
+# the potential gradient term to achieve a steady state
+# reference:
+# D. L. Williamson, J. B. Drake, J. J.HackRüdiger Jakob, P. N.Swarztrauber, (1992)
+# J Comp. Phys. 102 211-224
+
+# Constants of the Williamson2 test case
+const α  = π/4.0              # deviation of the coriolis term from zonal forcing
```
amartinhuertas, line 17: The same comment as in https://github.com/gridapapps/GridapGeosciences.jl/pull/18/files#r711833021 applies here for the `Williamsom2` initial conditions.

src/DiagnosticTools.jl
```diff
@@ -67,7 +67,7 @@ end
 """Wrapper to get a vector from a csv field. The fieldname argument should be passed as a symbol,
 i.e. :<fieldname>"""
 function get_scalar_field_from_csv(csv_file_path, fieldname)
-  t = CSV.read(csv_file_path, Table)
+  t = CSV.read(csv_file_path, DataFrame)
```
amartinhuertas, line 70: @cbladwell ... after replacing `Table` by `DataFrame`, do we still need the `TypedTables.jl` package in `Project.toml`?

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+    dt₁ = 2.0*dt
+  end
+
+  # multifield terms
+  u₁, h₁ = y₁
+  u₂, h₂ = y₂
+  uₘ, hₘ = yₘ
+
+  du₁,dh₁ = duh₁
+  du₂,dh₂ = duh₂
```
amartinhuertas, line 32: Remove this line (L32)

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  if leap_frog
+    dt₁ = 2.0*dt
+  end
+
+  # multifield terms
+  u₁, h₁ = y₁
+  u₂, h₂ = y₂
+  uₘ, hₘ = yₘ
+
+  du₁,dh₁ = duh₁
```
amartinhuertas, line 31: Move this line (L31) below, right before it is actually needed.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  du₂,dh₂ = duh₂
+
+  # 1.1: the mass flux
+  compute_mass_flux!(F,dΩ,V,RTMMchol,u₁*h₁)
+  # 1.2: the bernoulli function
+  compute_bernoulli_potential!(ϕ,dΩ,Q,L2MMchol,u₁⋅u₁,h₁,g)
+  # 1.3: the potential vorticity
+  compute_potential_vorticity!(q₁,H1h,H1hchol,dΩ,R,S,h₁,u₁,f,n)
+  # 1.4: assemble the velocity residual
+  bᵤ₁(v) = ∫(-1.0*(q₁ - τ*u₁⋅∇(q₁))*(v⋅⟂(F,n)))dΩ + ∫(DIV(v)*ϕ)dω
```
amartinhuertas, line 41: Move lines 41-46 to a function, say, `compute_velocity_depth_system_residual`. With reusability in mind (below), we can use an argument for the full operand that multiplies  `(v⋅⟂(F,n))`.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  get_free_dof_values(u₂) .= get_free_dof_values(uₘ) .+ dt₁ .* get_free_dof_values(du₁)
+  get_free_dof_values(h₂) .= get_free_dof_values(hₘ) .+ dt₁ .* get_free_dof_values(dh₁)
+
+  # 2.1: the mass flux
+  compute_mass_flux!(F,dΩ,V,RTMMchol,u₁*(2.0*h₁ + h₂)/6.0+u₂*(h₁ + 2.0*h₂)/6.0)
+  # 2.2: the bernoulli function
+  compute_bernoulli_potential!(ϕ,dΩ,Q,L2MMchol,(u₁⋅u₁ + u₁⋅u₂ + u₂⋅u₂)/3.0,0.5*(h₁ + h₂),g)
+  # 2.3: the potential vorticity
+  compute_potential_vorticity!(q₂,H1h,H1hchol,dΩ,R,S,h₂,u₂,f,n)
+  # 2.4: assemble the velocity residual
```
amartinhuertas, line 61: Transform lines 61-68 by a call to the function mentioned in https://github.com/gridapapps/GridapGeosciences.jl/pull/18/files#r711858624

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  bᵤ₂(v) = ∫(-0.5*(q₁ - τ*u₁⋅∇(q₁) + q₂ - τ*u₂⋅∇(q₂))*(v⋅⟂(F,n)))dΩ + ∫(DIV(v)*ϕ)dω
+  # 2.5: assemble the depth residual
+  bₕ₂(q) = ∫(-q*DIV(F))dω
+
+  # add A*[du₁,dh₁] to the [du₂,dh₂] vector
```
amartinhuertas, line 66: This comment should go right before line 70, right? Also, we are not adding, but substracting.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  bₕ₁(q) = ∫(-q*DIV(F))dω
+
+  bₕᵤ₁((v,q)) = bᵤ₁(v) + bₕ₁(q)
+  Gridap.FESpaces.assemble_vector!(bₕᵤ₁, get_free_dof_values(duh₁), Y)
+
+  # Solve for du₁, dh₁ over a MultiFieldFESpace
+  ldiv!(Blfchol, get_free_dof_values(duh₁))
+
+  # update
+  get_free_dof_values(u₂) .= get_free_dof_values(uₘ) .+ dt₁ .* get_free_dof_values(du₁)
```
amartinhuertas, line 52: Replace these two lines by `get_free_dof_values(y₂) .= get_free_dof_values(yₘ) .+ dt₁ .* get_free_dof_values(duh₁)`. 

I do not see why we have to split into different updates for each field (perhaps I am missing something).

If we merge both updates, this line https://github.com/gridapapps/GridapGeosciences.jl/pull/18/files#diff-e9f27dea57ae43bb4c320ef1dd2a44e8c9a3656344940df6cc4b4d4c032b4bf1R29 is not required anymore.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+
+  n = get_normal_vector(model)
+  dt₁ = dt
+  if leap_frog
+    dt₁ = 2.0*dt
+  end
+
+  # multifield terms
+  u₁, h₁ = y₁
+  u₂, h₂ = y₂
```
amartinhuertas, line 28: Move this line (L28) right before it is needed.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+
+  mul!(y_wrk, Amat, get_free_dof_values(duh₁))
+  get_free_dof_values(duh₂) .= get_free_dof_values(duh₂) .- y_wrk
+
+  # solve for du₂, dh₂
+  ldiv!(Bchol, get_free_dof_values(duh₂))
+
+  # update yⁿ⁺¹
+  du₂, dh₂ = duh₂
+  get_free_dof_values(u₂) .= get_free_dof_values(u₁) .+ dt .* get_free_dof_values(du₂)
```
amartinhuertas, line 78: Merge these two lines as in comment https://github.com/gridapapps/GridapGeosciences.jl/pull/18/files#r711860861

I think `du₂, dh₂ = duh₂` is not needed anymore.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  if leap_frog
+    lf = 2.0
+  end
+  Blf = M - lf*A
+  Blfchol = lu(Blf)
+
+  # multifield initial condtions
+  b₄((v,q)) = b₁(q) + b₂(v)
+  rhs2    = assemble_vector(b₄, Y)
+  yn      = FEFunction(Y, copy(rhs2))
```
amartinhuertas, line 168: Use `clone_fe_function`

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+
+  # build the potential vorticity lhs operator once just to initialise
+  bmm(a,b) = ∫(a*hn*b)dΩ
+  H1h      = assemble_matrix(bmm, R, S)
+  H1hchol  = lu(H1h)
+
+  function run_simulation(pvd=nothing)
+    diagnostics_file = joinpath(output_dir,"nswe__rosenbrock_diagnostics.csv")
+
+    clone_fe_function(space,f)=FEFunction(space,copy(get_free_dof_values(f)))
```
amartinhuertas, line 185: We are redefining clone_fe_function at several places. I would put it in `src`, e.g.,  a new `Helpers.jl` file and re-use it when needed. (I should have done that eariler, my fault).

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+                        h₀, u₀, f₀, g, H₀,
+                        λ, dt, τ, N; leap_frog=false,
+                        write_diagnostics=true,
+                        write_diagnostics_freq=1,
+                        dump_diagnostics_on_screen=true,
+                        write_solution=false,
+                        write_solution_freq=N/10,
+                        output_dir="nswe_eq_ncells_$(num_cells(model))_order_$(order)_rosenbrock")
+
+  # Forward integration of the shallow water equations
```
amartinhuertas, line 111: Might we factor out in functions some common operations among both solvers performed right before the time integration loop?

src/DiagnosticTools.jl
cbladwell: I have removed `TypedTables` from `Project.toml`

amartinhuertas: Ok, cool.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  # solve for du₂, dh₂
+  ldiv!(Bchol, get_free_dof_values(duh₂))
+
+  # update yⁿ⁺¹
+  du₂, dh₂ = duh₂
+  get_free_dof_values(u₂) .= get_free_dof_values(u₁) .+ dt .* get_free_dof_values(du₂)
+  get_free_dof_values(h₂) .= get_free_dof_values(h₁) .+ dt .* get_free_dof_values(dh₂)
+end
+
+function new_vtk_step(Ω,file,hn,un,wn)
```
amartinhuertas, line 82: I guess this function is common to both solvers, right?

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,252 @@
...
+  createvtk(Ω,
+            file,
+            cellfields=["hn"=>hn, "un"=>un, "wn"=>wn],
+            nsubcells=4)
+end
+
+function compute_mean_depth!(wrk, L2MM, h)
```
amartinhuertas, line 89: We are not currently calling this function. Why not?

src/ShallowWaterRosenbrock.jl
cbladwell: This assignment is no longer necessary and has been removed

src/ShallowWaterRosenbrock.jl
cbladwell: This has been removed as it is no longer necessary (along with L29)

src/ShallowWaterRosenbrock.jl
cbladwell: This comment has been updated to describe the sign change in `A` and moved as suggested

src/ShallowWaterRosenbrock.jl
cbladwell: Updated and additionally the assignments to `du₁`  and `dh₁` have been removed

src/ShallowWaterRosenbrock.jl
cbladwell: This has been moved to the second step

src/ShallowWaterRosenbrock.jl
cbladwell: Updated and `du₂, dh₂ = duh₂` removed

src/ShallowWaterRosenbrock.jl
cbladwell: Moved `clone_fe_function` from `ShallowWaterRosenbrock.jl` to `Helpers.jl`

src/ShallowWaterRosenbrock.jl
davelee2804: Hey @amartinhuertas, this is an optional function for computing the mean depth for use in the approximate Jacobian. Right now this is just being passed in has a constant, but I can definitely see scenarios where we have some initial state where we don't know the mean depth off hand. Is it OK if we keep it for future use?

driver/GalewskyShallowWaterExplicit.jl
davelee2804: done

test/Williamson2ShallowWaterRosenbrockTests.jl
davelee2804: done

src/ShallowWaterRosenbrock.jl
cbladwell: I don't think `clone_fe_function` can be used in L168 as `rhs2` is a vector

amartinhuertas: Ok, you are right. On the other hand, I think there is no need to `copy` `rhs2`. I would rewrite these lines as:

```
rhs2    = assemble_vector(b₄, Y)
ldiv!(Mchol, rhs2)
yn  = FEFunction(Y, rhs2)
```

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,243 @@
...
+  # compute the mean depth over the sphere, for use in the approximate Jacobian
+  mul!(wrk, L2MM, get_free_dof_values(h))
+  h_int = sum(wrk)
+  wrk  .= 1.0
```
amartinhuertas, line 86: Replace the next three lines by `a_int=sum(A,dims=[1,2])`

src/ShallowWaterRosenbrock.jl
amartinhuertas: Ok, let us keep it here, but see my comment below.

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,243 @@
+function shallow_water_rosenbrock_time_step!(
+     y₂, ϕ, F, q₁, q₂, duh₁, duh₂, H1h, H1hchol, y_wrk,  # in/out args
```
amartinhuertas, line 2: If I am not wrong, `y2` represents `y^{n+1}` in the draft paper, and `y1` represents  `y^{n}`, right? What does `ym` represent in the draft paper? As far as I can see, the calling function is providing `y^{n-1}` as an argument to the `ym` function parameter. Is that we we actually want/what it is written in the draft paper? Can we use better parameter names to reflect what `y2` `y1` and `ym` represent using draft notation?

src/ShallowWaterRosenbrock.jl
```diff
@@ -0,0 +1,243 @@
...
+      compute_diagnostic_vorticity!(wn, dΩ, S, H1MMchol, un, get_normal_vector(model))
+      dump_diagnostics_shallow_water!(h_tmp, w_tmp,
+                                      model, dΩ, dω, S, L2MM, H1MM,
+                                      hn, un, wn, ϕ, F, g, istep, dt,
+                                      diagnostics_file,
+                                      dump_diagnostics_on_screen)
+    end
+    # time step iteration loop
+    for istep in 2:N
+      get_free_dof_values(ym2) .= get_free_dof_values(ym1)
```
amartinhuertas, line 211: I do not see why these two copies are necessary. If am not wrong, I think we can achieve the same effect using the following line, which relies on pointer association:

`ym2,ym1,yn=ym1,yn,ym2`  

BTW ... whay does m2 and m1 denote? minus 2 and minus 1?

davelee2804: Yes correct, minus 1 and minus 2. I tried this previously and that is when we had the three repeating time step problem. Could be I did something else wrong but the problem was fixed by reverting to the current approach

amartinhuertas: Yes, indeed, I was aware of that. In particular, you modified this in this commit:

https://github.com/gridapapps/GridapGeosciences.jl/pull/18/commits/0e751edc86fce4a029790b83de5985aae8fc4395

which, as you say, before you had exactly what I am proposing here. 

I have been taking a very careful look at the code, step by step, and I do no see how this line cannot work. Perhaps I can try to use the debugger to understand what might be going on.

davelee2804: I was wondering if maybe the pointer swapping does not work for multifields for some reason. Probably it is something silly I did, I will double check as well...

src/ShallowWaterRosenbrock.jl
cbladwell: Good point. I have updated this now

src/ShallowWaterRosenbrock.jl
davelee2804: Indeed, good point. Have migrated this into Helpers.jl

src/ShallowWaterRosenbrock.jl
davelee2804: done, thanks

src/ShallowWaterRosenbrock.jl
davelee2804: Have moved the function space initialisation and mass matrix assembly and LU factorisation into separate functions in Helpers, and using these in both solvers

src/ShallowWaterRosenbrock.jl
amartinhuertas: @davelee2804 ... FYI ... I have solved this in commit d9a0cca076e72e7d6c583f4538099071c631a568

We had two errors: 
1. If we assign `ym2` to `ym1` without storing temporarily the object `ym2` points to into an auxiliary variable, we end up assigning `yn` to `ym1` again ... (this is 1st year introduction to computer programming hehehe my fault as well)
2. We have to reassign `un,hn` at each iteration as the object `yn` points to changes at each iteration as well.

amartinhuertas: BTW ... can 1. affect as well to the explicit solver implementation?

src/Helpers.jl
```diff
@@ -0,0 +1,35 @@
...
+  Q = FESpace(model, reffe_lgn; conformity=:L2)
+  P = TrialFESpace(Q)
+  reffe_lgn = ReferenceFE(lagrangian, Float64, order+1)
+  S = FESpace(model, reffe_lgn; conformity=:H1)
+  R = TrialFESpace(S)
+
+  R, S, U, V, P, Q
+end
+
+function setup_mass_matrices(dΩ, R, S, U, V, P, Q)
```
amartinhuertas, line 18: Rename `setup_mass_matrices` by `setup_and_factorize_mass_matrices`

src/ShallowWaterRosenbrock.jl
davelee2804: Thanks for addressing thing @amartinhuertas . I would have thought 1 would indeed affect the explicit solver. I suppose it depends on the order of the pointer re-assignment. When `un`, `hn` are re-assinged in the explicit solver we don't actually need them, it is OK if they are junk when they are passed into the solver. I don't know the order of assignment in `um2,um1,un = um1,un,um2`, but if `un` is effectively the `aux` variable then I suppose we are ok...

davelee2804: ...Just re-checked the results of the explicit solver. There is no three-peat in the diagnostic output

amartinhuertas: > I don't know the order of assignment in um2,um1,un = um1,un,um2

This syntax expression is equivalent to 

```
um2 = um1
um1 = un
un    =  um2
```

i.e, left to right order. 

>  I would have thought 1 would indeed affect the explicit solver. I suppose it depends on the order of the pointer re-assignment. When un, hn are re-assinged in the explicit solver we don't actually need them, it is OK if they are junk when they are passed into the solver. 

The issue is that we have pointer aliasing (if we dont use the `aux` variable trick), i.e., `um2` and `un` point to the same object. Besides, we lose track of the object that `um2` initially points to. (This would result in memory leaks in languages without GC). I think this is dirty/prone to errors, even if we have a working code. As a consequence of pointer aliasing, if `un` gets modified prior to `um1` is read, the code wont work. I guess this is not the case within the explicit time step function. (I did not check), thus explaining why it is working. I would use `aux` better to avoid potential future trouble with this.

src/GridapGeosciences.jl
```diff
@@ -28,7 +30,11 @@ module GridapGeosciences
   export Eₖ, Eₚ, Eₜ
   export compute_kin_to_pot!, compute_pot_to_kin!, compute_total_mass!
   export dump_diagnostics_shallow_water!
+  export compute_mass_flux!, compute_bernoulli_potential!, compute_potential_vorticity!
   export shallow_water_explicit_time_step!
-  export shallow_water_time_stepper
+  export shallow_water_explicit_time_stepper
+  export shallow_water_rosenbrock_time_step!
```
amartinhuertas, line 36: Do we need to export `shallow_water_rosenbrock_time_step!` ? If yes, why?

The same applies to `shallow_water_explicit_time_step!

amartinhuertas: what about the rest of symbols? Which ones need to be actually exported?

davelee2804: No, I don't think we do, this was a legacy from before the time steppers functions were implemented. I'll fix and look for other symbols we don't need to export...

src/ShallowWaterRosenbrock.jl
davelee2804: Personally I am OK with having these as `y2`, `y1`, `ym`, but if you prefer that they be something like `yp2`, `yp1`, `y0` or something like that I can change them. I just don't like having too many underscore letters in a row

davelee2804: Sounds good, done :)

src/ShallowWaterRosenbrock.jl
davelee2804: I've updated the explicit solver to include `u_aux`, `h_aux` in the pointer cycling in this commit here: bdeb9343d2c15479cbb52a4eb40e8b63e0afe6a5

PR #21, opened 2021-09-22
Shallow water imex solver

amartinhuertas:
Work in progress ... opening PR to start discussion as needed.


davelee2804:
Hi @amartinhuertas , As well as the williamson2 test case, I've run the imex solver through the galewsky test case. Results are as expected
vort_day6_imex

vort_cons

mass_cons
12-99cb-5e86a8704f05.png)
ener_cons
d:

There is one optimisation issue I know of: L46-47, I'm sure there is a way to do these matrix operations in place using preallocated matrices, I just don't know how to do it with Julia's LinearAlgebra library, and couldn't find any documentation on this, do you know? Otherwise I think this is ready for review....

amartinhuertas:
Ok, thanks for your work @davelee2804 !

> There is one optimisation issue I know of: L46-47, I'm sure there is a way to do these matrix operations in place using preallocated matrices, I just don't know how to do it with Julia's LinearAlgebra library, and couldn't find any documentation on this, do you know? Otherwise I think this is ready for review....

Ok. Let me take a look at that and I will come back to you.

src/ShallowWaterIMEX.jl
```diff
@@ -0,0 +1,216 @@
+function shallow_water_imex_time_step!(
+     h₂, u₂, uₚ, ϕ, F, q₁, q₂,                               # in/out args
+     H1h, H1hchol, h_wrk, u_wrk, A_wrk, Bchol,               # more in/out args
+     model, dΩ, dω, V, P, Q, R, S, f, g, h₁, u₁, u₀,         # in args
+     RTMMchol, L2MMchol, RTMM, L2MMinvD, dt, τ, leap_frog)   # more in args
+
+  # energetically balanced implicit-explicit second order shallow water solver
+  # reference: eqns (31-33) of
+  # https://github.com/BOM-Monash-Collaborations/articles/blob/main/energetically_balanced_time_integration/EnergeticallyBalancedTimeIntegration_SW.tex
+  #
```
amartinhuertas, line 10: The signature description is outdated. I think the same applies to the other functions. I would either update (if we think they are strictly necessary) or remove them (if we think that their names are sufficiently meaning revealing).

src/ShallowWaterIMEX.jl
```diff
@@ -0,0 +1,218 @@
...
+  # dΩ         : measure of the elements
+
+  n = get_normal_vector(model)
+  # explicit step for provisional velocity, uₚ
+  dt1 = dt
+  if leap_frog
+    dt1 = 2.0*dt
+  end
+
+  Fv,h₁v,h₂v = get_free_dof_values(F,h₁,h₂)
```
amartinhuertas, line 30: @cbladwell @davelee2804 ... note that I have introduced a new method for the multi-method `get_free_dof_values` in `Helpers.jl`. Using this method, we just call `get_free_dof_values` once in the preamble. 

@cbladwell ... can you please give us a hand and apply the same pattern in the other two solvers as per required? 

Thanks!

src/ShallowWaterIMEX.jl
```diff
@@ -0,0 +1,218 @@
...
+  # 1.4: solve for the provisional velocity
+  compute_velocity!(uₚ,dΩ,dω,V,RTMMchol,u₀,q₁-τ*u₁⋅∇(q₁),F,ϕ,n,dt1,dt1)
+
+  # solve for the second order, semi-implicit mass flux
+  # 2.1: mass flux component using the previous depth (explicit)
+  compute_mass_flux!(F,dΩ,V,RTMMchol,h₁*(2.0*u₁ + uₚ)/6.0)
+  # 2.2: mass flux component using the current depth (implicit)
+  adv(p,v) = ∫(v⋅((u₁ + 2.0*uₚ)/6.0)*p)dΩ
+  Gridap.FESpaces.assemble_matrix!(adv, A_wrk, P, V)
+  # TODO: there must be a way to do these two operations in place using preallocated matrices!...
```
amartinhuertas, line 47: Ok, after looking at  the full structure of the implicit compound operator, I think I now know which is the the efficient way to go to implement these operations. Namely:

1. We do not assemble `L2MMinv` nor `D` into global matrices (as we are doing now), but keep them sub-assembled permanently into cell-wise arrays with the cell contributions. In both cases, we transform the corresponding `LazyArrays`s into plain Julia arrays using `collect` (as we are doing now with  `L2MMinv` prior global assembly, only temporarily, though).
2. We perform the `L2MMinv` * `D` operation cell-wise, and keep it in a permanently stored Julia array.
3. We do not assemble  `adv`  into `A_wrk`. Instead, we obtain the `LazyArray` corresponding to `adv(p,v)`, and build a new `LazyArray` encoding the `RTMM + dt*A` operation at the cell level. In order to do so, we need 2. and the cell-wise subassembled vector corresponding to `RTMM`, which once we have computed its LU factors, I think is no longer needed in fully assembled form (if am not missing something). 
4. We assemble 3.  using U,V spaces. 

Note that, using this approach, we do not need to perform any global sparse-matrix to sparse-matrix product. Of course, this is possible given the decoupled structure of the L2 space DOFs.

Does that sound reasonable?

davelee2804: Hey @amartinhuertas , my only concern is that `D` is not block diagonal. It has a divergence operator in there, so the sparsity pattern with traverse DOFS from multiple elements. Is what you propose still OK for a non-block-diagonal `D`?

amartinhuertas: > Hey @amartinhuertas , my only concern is that D is not block diagonal. It has a divergence operator in there, so the sparsity pattern with traverse DOFS from multiple elements. Is what you propose still OK for a non-block-diagonal D?

Yes, you are right, it is not block diagonal. However, still it has a particular structure that enables this optimization. Namely,  it can be split into blocks of rows (as many blocks as cells, as many rows in each block as L2 DoFs in each cell), where each block of rows can be computed independently of each other locally at each cell. Do u agree?

davelee2804: Yes, I agree. But those blocks will not necessarily be adjacent in the sparse matrix (or memory). Is that a problem?

amartinhuertas: I am not sure I understand. Keeping the entries of D into (small) rectangular dense matrices and operating among these using the level 3 BLAS is (much) more efficient than assembling them and then performing the operation globally with the sparse matrices (essentially, level 1 BLAS, integer memory accesses to retrieve column/row identifiers, etc.).

davelee2804: OK, sorry, I think it was I who misunderstood. You are completely ignoring the full sparse matrix, and just carrying around the "patches" of matrix entries (which will be thin and tall for divergence, since the gradient is applied in the column; the trial function) then using BLAS 3 for these small (dense) matrices. OK, that makes sense. Sorry for the confusion.

amartinhuertas: Yes exactly. The full global operation can be decoupled into cell wise local operations + global final assembly. 

I think I am not missing anything ... But one never knows ... Best to subject ideas to other brains :-)

davelee2804: No, I think it is a good idea. I was just mis-interpreting. You are way ahead of me :)

amartinhuertas: Ok. Do u want to give it a go? Or should I proceed instead? Whatever your prefer ...

src/ShallowWaterIMEX.jl
davelee2804: Good point, I've removed them.

src/ShallowWaterIMEX.jl
davelee2804: The bit I'm unsure of is how to assemble the Gridap bilinear form into an array of dense matrices. I'm happy to try, but in all honesty I don't want to invest too much time in this if it isn't straight forward. I think long term (thermal shallow water and beyond) the explicit and Rosenbrock solvers are probably the way forward. Once we have a Krylov solver the longer time steps of the Rosenbrock solver should give us a performance boost, and between that and the explicit solver I think we are covered. The IMEX solver has the added complexity of a semi-implicit solver, and none of the performance benefit of larger time steps. It is a good comparison for the paper I think, but I'm not sure it is of much long term value. If I was going to invest a lot of time in low level work, the thing I would want to look at is downwinding of trial functions for the potential vorticity equation. This is what I do in MiMSEM, and I think maybe it is why MiMSEM can run with neutral stability. The downwinded trial functions allow for higher order dispersion, while the APVM does not see the order of the basis functions, which is why I think we need the added damping of the L-stable Rosenbrock method...

amartinhuertas: Ok, no worries. I understand what you are saying. It does not make much sense to optimize a method which is not likely to become the method of choice. Anywat, as I was the one suggesting this approach, it makes sense that I implement it. I think I can implement it in moderate time.

davelee2804: OK @amartinhuertas , thanks for this. If it ends up being a lot of work don't worry too much. The current performance is adequate to generate results for the paper. I think we are nearly ready to write up all the shallow water work, what do you think?

amartinhuertas: > I think we are nearly ready to write up all the shallow water work, what do you think?

Let us discuss this in a separate issue at the article repo, better.

amartinhuertas: > OK @amartinhuertas , thanks for this. If it ends up being a lot of work don't worry too much.

I already implemented this in commit a4d18aa3c94df4916a6d617f81303fed3f16c1c9. The L2 errors are as in your version (i.e., the tests pass). However, the second column (from the left) of the diagonostics (I think it is the potential vorticity, right?), shows some sensitivity to rounding errors. Is this as expected?

I have taken some  time measurements in my laptop (nothing systematic, though). In terms of time, both solutions are comparable for `order=1` (RT1-DG1-CG2-RT1), however, as I would expect, as we go to higher degrees, e.g., `order=3`, the new version is faster (the larger the order, the higher performance of the LEVEL 3 BLAS). However, there is not such a big deal, as the steps that were optimized do not concentrate a significant portion of the total computation time of the IMEX iteration (see attached files for more details). In numbers (`order=3`; 40x40 cells/panel): 

#### PRIOR TO OPTIMIZATION 

```
  @time A        = A_wrk*L2MMinvD
  @time B        = RTMM + dt*A
  @time mul!(h_wrk, L2MMinvD, Fv)
  @time mul!(u_wrk, A_wrk, h_wrk)
```

```
  0.492661 seconds (8 allocations: 4.026 GiB, 2.37% gc time)     # A_wrk*L2MMinvD
  0.760396 seconds (12 allocations: 693.751 MiB, 66.03% gc time) # RTMM + dt*A
  0.007396 seconds (1 allocation: 48 bytes)
  0.008844 seconds
```

#### AFTER OPTIMIZATION 

```
@time assemble_implicit_compound_operator!(B,P,U,V,dt,dΩ,A,RTMM,L2MMinvD) # B = RTMM + dt*A*L2MMinv*D
@time assemble_Asub_mul_u!(h_wrk,U,P,dΩ,L2MMinvD,Fv)                                    # h_wrk = L2MMinvD * Fv
@time assemble_Asub_mul_u!(u_wrk, P, U, dΩ, A, h_wrk)                                        # u_wrk = A * h_wrk
```

```
0.325012 seconds (164 allocations: 153.531 KiB) # B = RTMM + dt*A*L2MMinv*D
0.007599 seconds (147 allocations: 16.078 KiB)  # h_wrk = L2MMinvD * Fv
0.007359 seconds (144 allocations: 15.125 KiB) # u_wrk = A * h_wrk
```

**However**, note there is a brutal difference in memory consumption among both solutions. 


   
[imex_dave.txt](https://github.com/gridapapps/GridapGeosciences.jl/files/7229063/imex_dave.txt)
[imex_optimized.txt](https://github.com/gridapapps/GridapGeosciences.jl/files/7229064/imex_optimized.txt)

src/ShallowWaterIMEX.jl
amartinhuertas: Addressed here ad29e55e1673461bdba7859ddaa294a4594e9d51

amartinhuertas:
Hi  @davelee2804 ! The IMEX solver PR is ready to be accepted from my side (subject to passing tests). Thanks for your work!

davelee2804:
Thanks for all your great work here @amartinhuertas ! I will run your changes with the Galewsky test case just to make doubly sure everything is OK (I'm sure it is).

davelee2804:
Hi @amartinhuertas , I ran your changes through the Galewsky test case. Everything looks good. Thanks again for your help here! If you are happy I am happy to merge the PR, what do you think? Are we still missing anything?

amartinhuertas:
> Hi @amartinhuertas , I ran your changes through the Galewsky test case. Everything looks good. Thanks again for your help here! If you are happy I am happy to merge the PR, what do you think? Are we still missing anything?

Ok, great! I think it can be accepted. Accepting ...

PR #22, opened 2021-09-29
Theta method full newton nswe solver

amartinhuertas:
**Reopening this work as a different PR. The old PR cannot be merged into `master` (due to some mess that I did yesterday in the `master` branch, but that is now solved).**

Hi @davelee2804 @cbladwell,

in this PR I have written a "skeleton" code (it is just a partial implementation, to be completed/validated) of a NSWE solver using Gridap's high level API. 

* It uses the APVM method for spatial discretization, i.e., Eqs. (70)-(73) in [Rognes et. al](https://gmd.copernicus.org/articles/6/2099/2013/) and the theta method implicit method for time discretization (the fully discrete nonlinear problem to be solved at each time step is in particular avaliable in Eqs. (75)-(78) of that paper). 
* You can see the evaluation of this nonlinear operator (i.e., residual) coded here: https://github.com/gridapapps/GridapGeosciences.jl/commit/88418fe8cdf327412822361733626dd4988056db#diff-3d44a05a1e12bbf492c4d4eea9a151784def9fe3a3c34312c3e9a5f35e106ff6R208
where the first argument is the current evaluation point, and the latter a basis of the monolothic system test functions.

However, some caveats:
 1. I could **NOT** compute the Jacobian of this operator using AD (because of a known issue in Gridap 0.16 that has to be solved; see https://github.com/gridap/Gridap.jl/issues/584 for details) We will need an approximate Jacobian, or to resort to a different discretization, e.g., the one at the end of Section 5.5 of Rognes paper (this discretization is not energy conserving, though). 
 2. The JIT compilation of the code triggered from `res` down the tree takes forever with Julia 1.6.2 (it seems to enter in a infinite loop, although I could not confirm it). With Julia 1.5.4 the issue disappears. Again, I think similar issues have appeared with Julia 1.6+Gridap in similar contexts. We should investigate this further. 

Also I wonder the following:
- Can we actually exploit the DIV+reference integration for this term: `(∇⋅(v))*(0.5*ui⋅ui)` ?
- Is the current (partial) implementation correct? (as I could not validate it yet) 

@davelee2804 @cbladwell perhaps you can contribute and complete it (essentially implement a crude approximation of the Jacobian as @davelee2804 mentioned). What do u think?  I think that this driver shows that we have all the ingredients in Gridap required to solve the NSWEs at the discretization level (e.g., I could evaluate the residual as implemented in `res`).  I am happy with you completing the current method, or to consider another method as well (e.g., one of Cotter's book for the NSWE different from the APVM). 

Please, once you start modifying the driver, can you please open a branch and PR to this branch so that I can help you and review etc. ? Thanks!

## Tasks 

- [x] Complete the exact Jacobian, i.e., complete the `jac` function definition statement. https://github.com/gridapapps/GridapGeosciences.jl/blob/11fadb3caad4585c5e30526d0db58d865ad0bdb6/test/ThetaMethodFullNewtonNSWECubedSphereTests.jl#L171
- [x] Implement/Complete diagnostics (currently commented out from WaveEquationSolver)
- [ ] Validate the solver against different benchmarks from Williamson et. al. 
- [x] Solve the following question: Can we actually exploit the DIV+reference integration for this term: `(∇⋅(v))*(0.5*ui⋅ui)` ? (see PR https://github.com/gridapapps/GridapGeosciences.jl/pull/12 for more details)
- [x] Investigate issues with AD of the fully coupled system. (no longer needed, we have the manual Jacobian).
- [ ] **UPDATE**: *by a more clever ad-hoc implementation of the Backward Euler method I could reduce first JIT compilation times to around 10 mins with Julia 1.5.4. Not sure with Julia 1.6.2., we will see in Github actions.* Investigate JIT compilation issues with Julia 1.6.2. Switch to Julia 1.5.4 in the meantime.


davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L96

I'm not sure that `dt/2` is always the best value for `\tau`. In my experiments with the Rosenbrock solver I found that a shorter value is sometimes better. I would pass this in as a parameter rather than hard code it.

davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L107

This is a style comment (but a major one), and I leave it up to you to make the decision as to weather to do this or not, but I would express the solution variables as `un, hn, uk, hk`, not `un, hn, \Delta u, \Delta h`, and then solve for `\Delta uk, \Delta hk`, with `uk = uk + \Delta uk, hk = hk + \Delta hk`. I know that what you are doing is consistent with what Rognes et al do, and that this works for the \theta method, but it is not a very helpful formulation for the Crank-Nicolson or energetically balanced schemes, which represent the temporal cross terms differently. Again, your decision here....

davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L151
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L160
The vorticity is being diagnosed for both the diagnostics and the field dumps. Most likely this will mean it is being computed twice. Perhaps a boolean could be added to the write_solution call to ensure this isn't called twice

davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L7
Can we just use `compute_mass_flux` from the explicit solver here?

davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L14
Can we just use `diagnose_potential_vorticity` from the explicit solver instead?

davelee2804:
https://github.com/gridapapps/GridapGeosciences.jl/blob/theta_method_full_newton_nswe_solver/src/ShallowWaterThetaMethodFullNewton.jl#L2
Should this be moved into the initial conditions file?

amartinhuertas:
Hi @davelee2804 ! Thanks for your insightful comments! I will address them carefuly and come back to you. Note that the most appropriate way to add comments in a PR (next time) is directly in the line of code in which you want to make your point, in the "Files changed" tab of the PR page.

amartinhuertas:
In the "Files changed" tab of the PR page, and sitting on a given line of code, you have to click on the + sign which is popped up when you pass the mouse pointer in between the line of code identifier and the line of code itself.

davelee2804:
Thanks @amartinhuertas , I was wondering about that. will do

amartinhuertas:
Hi @davelee2804 ! I have addresed all your comments except the one that involves more work/more careful thinking (i.e., https://github.com/gridapapps/GridapGeosciences.jl/pull/22#issuecomment-932673370). I am happy to do this work as well, although I prefer to devote some more time to understand. (also our conversations in slack etc.) 

In regards to:

https://github.com/gridapapps/GridapGeosciences.jl/pull/22#issuecomment-932673867

note that I have also done it. BUT ... the same can be applied to the other time steppers (if I am not wrong). Can you apply the patch there as well?

On the other hand, and related to the previous, I have the impression that we are replicating a very similar pattern in all time steppers (this is the cause we have to apply the patch in all places). It is not a massive amount of code the one we are replicating (basically control of the time integration loop), but perhaps we may think a little bit more how can we abstract the time integration loop into a  structure/pattern common to all solvers (if such possibility exists/it is convenient)

davelee2804:
Hey @amartinhuertas , thanks for all your work here, the code is looking great, very succinct. With regard to https://github.com/gridapapps/GridapGeosciences.jl/pull/22#issuecomment-932673370, what is your preference? You have done a lot of great work here, and I'm happy to merge in the PR as is. We could then look at re-formulating the nonlinear terms in the residuals as future work (it is called a 'ThetaMethod' after all). What do you preference? One final question: what is the energy conservation error? The explicit and IMEX integrators loose O(1.0e-9) over the simulation, the Rosenbrock a bit more (on account of the approximate Jacobian). What is the loss here? Also can you run time centered with the super long time step (9600s), or does this require a shorter time step?

amartinhuertas:
> Hey @amartinhuertas , thanks for all your work here, the code is looking great, very succinct. With regard to #22 (comment), what is your preference? You have done a lot of great work here, and I'm happy to merge in the PR as is. We could then look at re-formulating the nonlinear terms in the residuals as future work (it is called a 'ThetaMethod' after all). Otherwise I can help with the re-formulation of the nonlinear terms as Crank-Nicolson and/or energetically balanced. I'm happy either way.

Let me think about this a little bit more and then we decide. 

> One final question: what is the energy conservation error? The explicit and IMEX integrators loose O(1.0e-9) over the simulation, the Rosenbrock a bit more (on account of the approximate Jacobian). What is the loss here? 

For BE and dt=9600s I shared the full diagnostics file in Slack ...  can you generate the plots? (note that i could not run the full 20 days within the 48 hours wall clock time). 

> Also can you run time centered with the super long time step (9600s), or does this require a shorter time step?

I tried Trapezoidal Rule with dt=9600s, and Newton for the first time step diverged: 

```
Iter     f(x) inf-norm    Step 2-norm 
------   --------------   --------------
     0     1.501067e+11              NaN
     1     1.105992e+25     4.185006e+37
     2     4.350110e+28     5.400823e+42
     3     1.553421e+26     5.486737e+42
     4     6.189773e+26     6.446322e+41
     5     1.779207e+28     1.368093e+43
     6     2.114128e+30     3.841482e+45
     7     8.529852e+31     4.038874e+46
     8     4.711679e+39     1.021888e+52
     .....
    23              NaN    3.825513e+264
```

I have reduced dt to 4800 (i.e, halved dt), and seems to be running ... will share when it finishes.

davelee2804:
Hey @amartinhuertas , Sorry, I wasn't very clear, I was intending to ask about the energy conservation for the time centred scheme (it was late and I wasn't thinking properly). Here is the plot for the backwards euler energy conservation. 
backwards_euler_en_cons
This is not dissimilar to the Rosenbrock energy conservation error, and is probably to be expected given then excessive damping of the BE scheme that @santiagobadia mentioned. I'm impressed that the time centred scheme is running with a time step of 4800s. Also (separate note), love that you added the vortex roll up movie to the geosciences page. I could look at that thing for hours :)

amartinhuertas:
> The JIT compilation of the code triggered from res down the tree takes forever with Julia 1.6.2 (it seems to enter in a infinite loop, although I could not confirm it). With Julia 1.5.4 the issue disappears. Again, I think similar issues have appeared with Julia 1.6+Gridap in similar contexts. We should investigate this further.

A follow up on this very annoying issue. Together with Julia compiler experts, we could isolate this issue, and narrow down where it does come from. To access to a workaround we can temporaritly use the Gridap.jl branch compilefast .  I.e.,  by typing `pkg> add Gridap#compilefast`. Alternatively, we can downgrade compiler optimizations to `O1` running julia as `julia -O1`. 
See also https://github.com/gridap/Gridap.jl#known-issues

PR #23, opened 2021-10-01
Tswe exp

davelee2804:
I have added an explicit solver for the thermal shallow water equations, as well as a default test configuration with a constant buoyancy (of value g) for which the thermal shallow water equations default to the regular shallow water equations. Under these conditions the solver replicates the results of the Galewsky test case, with the buoyancy staying constant
vort_day6_g
buoy_day6_g
ener_cons
mass_cons
vort_cons

I then modified the initial condition so that the initial perturbation is all in the depth field, so that the initial buoyancy deviates from the constant state. The resulting buoyancy perturbation is advected with the flow field, generating high frequency oscillations
vort_day6_b
buoy_day6_b

### DIFF
--- a/src/DiagnosticTools.jl
+++ b/src/DiagnosticTools.jl
@@ -117,7 +117,7 @@ function dump_diagnostics_thermal_shallow_water!(h_tmp, w_tmp,
   kin_i     = Eₖ(u,h,dΩ)
   int_i     = Eᵢ(h,E,dΩ)
   pow_k2p_i = sum(∫(ϕ*DIV(F))dω)
-  pow_k2i_i = 0.5*sum(∫h*DIV(eF)dω)
+  pow_k2i_i = 0.5*sum(∫(h*DIV(eF))dω)
 
   append_to_csv(output_file;
                 time       = step*dt/24/60/60,
--- a/src/DiagnosticTools.jl
+++ b/src/DiagnosticTools.jl
@@ -13,6 +13,13 @@ function Eₚ(hh,g,dΩ)
 end
 
 """
+  Internal energy (thermal shallow water)
+"""
+function Eᵢ(hh,Eh,dΩ)
+  0.5*sum(∫(hh*Eh)dΩ)
+end
+
+"""
   Total energy
 """
 function Eₜ(uh,H,hh,g,dΩ)
@@ -97,5 +104,32 @@ function dump_diagnostics_shallow_water!(h_tmp, w_tmp,
   if dump_on_screen
     @printf("%5d %14.9e %14.9e %14.9e %14.9e %14.9e %14.9e\n",
              step, mass_i, vort_i, kin_i, pot_i, kin_i+pot_i, pow_i)
+  end
+end
+
+function dump_diagnostics_thermal_shallow_water!(h_tmp, w_tmp,
+                                         model, dΩ, dω, S, L2MM, H1MM,
+                                         h, u, E, w, ϕ, F, eF, step, dt,
+                                         output_file, dump_on_screen)
+
+  mass_i    = compute_total_mass!(h_tmp, L2MM, get_free_dof_values(h))
+  vort_i    = compute_total_mass!(w_tmp, H1MM, get_free_dof_values(w))
+  kin_i     = Eₖ(u,h,dΩ)
+  int_i     = Eᵢ(h,E,dΩ)
+  pow_k2p_i = sum(∫(ϕ*DIV(F))dω)
+  pow_k2i_i = 0.5*sum(∫h*DIV(eF)dω)
+
+  append_to_csv(output_file;
+                time       = step*dt/24/60/60,
+                mass       = mass_i,
+                vorticity  = vort_i,
+                kinetic    = kin_i,
+                internal   = int_i,
+                power_k2p  = pow_k2p_i,
+                power_k2i  = pow_k2i_i)
+
+  if dump_on_screen
+    @printf("%5d %14.9e %14.9e %14.9e %14.9e %14.9e %14.9e %14.9e\n",
+             step, mass_i, vort_i, kin_i, int_i, kin_i+int_i, pow_k2p_i, pow_k2i_i)
   end
 end
--- a/src/Helpers.jl
+++ b/src/Helpers.jl
@@ -41,3 +41,10 @@ function new_vtk_step(Ω,file,hn,un,wn)
             cellfields=["hn"=>hn, "un"=>un, "wn"=>wn],
             nsubcells=4)
 end
+
+function new_vtk_step_twse(Ω,file,hn,un,wn,en)
+  createvtk(Ω,
+            file,
+            cellfields=["hn"=>hn, "un"=>un, "wn"=>wn, "en"=>en],
+            nsubcells=4)
+end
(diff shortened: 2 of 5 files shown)


src/Helpers.jl
```diff
@@ -41,3 +41,10 @@ function new_vtk_step(Ω,file,hn,un,wn)
             cellfields=["hn"=>hn, "un"=>un, "wn"=>wn],
             nsubcells=4)
 end
+
+function new_vtk_step_tswe(Ω,file,hn,un,wn,en)
```
amartinhuertas, line 45: I think we can do better here. I would try to stick into a single `new_vtk_step` function. 

To this end, you can modify the current signature to (Omega, file, cellfields) and pass directly, e.g., `["hn"=>hn, "un"=>un, "wn"=>wn]` as an argument to the last parameter.

src/DiagnosticTools.jl
```diff
@@ -99,3 +106,30 @@ function dump_diagnostics_shallow_water!(h_tmp, w_tmp,
              step, mass_i, vort_i, kin_i, pot_i, kin_i+pot_i, pow_i)
   end
 end
+
+function dump_diagnostics_thermal_shallow_water!(h_tmp, w_tmp,
+                                         model, dΩ, dω, S, L2MM, H1MM,
+                                         h, u, E, w, ϕ, F, eF, step, dt,
+                                         output_file, dump_on_screen)
+
+  mass_i    = compute_total_mass!(h_tmp, L2MM, get_free_dof_values(h))
```
amartinhuertas, line 115: I would wrap around lines 115-119 in a function, say, `compute_diagnostics_shallow_water!`, that can be called both from `dump_diagnostics_thermal_shallow_water!` and `dump_diagnostics_shallow_water!`

src/ThermalShallowWaterExplicit.jl
```diff
@@ -0,0 +1,219 @@
...
+  Gridap.FESpaces.assemble_vector!(b, get_free_dof_values(u1), V)
+  ldiv!(RTMMchol, get_free_dof_values(u1))
+end
+
+function thermal_shallow_water_explicit_time_step!(
+     h₂, u₂, E₂, hₚ, uₚ, Eₚ, ϕ, F, q₁, q₂, e₁, e₂, H1h, H1hchol, dT,  # in/out args
+     model, dΩ, dω, V, Q, R, S, f, h₁, u₁, E₁, hₘ, uₘ, Eₘ,            # in args
+     RTMMchol, L2MMchol, dt, τ, leap_frog)                            # more in args
+
+  # energetically balanced explicit second order thermal shallow water solver
```
amartinhuertas, line 31: Is this comment up to date/comprehensive ?

src/ThermalShallowWaterExplicit.jl
```diff
@@ -0,0 +1,219 @@
...
+  # Forward integration of the shallow water equations
+  Ω = Triangulation(model)
+  dΩ = Measure(Ω, degree)
+  dω = Measure(Ω, degree, ReferenceDomain())
+
+  # Setup the trial and test spaces
+  R, S, U, V, P, Q = setup_mixed_spaces(model, order)
+
+  # assemble the mass matrices
+  H1MM, RTMM, L2MM, H1MMchol, RTMMchol, L2MMchol = setup_and_factorize_mass_matrices(dΩ, R, S, U, V, P, Q)
```
amartinhuertas, line 102: I keep seeing the same pattern. `RTMM` is not needed outside the function, only inside for computing `RTMMchol`. Do we need to keep `RTMM` till the end of the simulation?

src/ThermalShallowWaterExplicit.jl
```diff
@@ -0,0 +1,219 @@
...
+
+  # Project the initial conditions onto the trial spaces
```
amartinhuertas, line 104: Arent we repeating in lines 104-123? It seems that the four blocks of code share the same pattern. Perhaps we can wrap around one block in a function, and call it repeteadly. Does this comment apply to the other solvers?

src/DiagnosticTools.jl
```diff
@@ -12,6 +12,13 @@ function Eₚ(hh,g,dΩ)
   0.5*g*sum(∫(hh*hh)dΩ)
```
amartinhuertas, line 12: can Epi call Ei?

src/ThermalShallowWaterExplicit.jl
```diff
@@ -0,0 +1,219 @@
...
+
+  n = get_normal_vector(model)
+  # explicit step for provisional velocity, uₚ
+  dt1 = dt
+  if leap_frog
+    dt1 = 2.0*dt
+  end
+
+  # 1.1: the mass flux
+  compute_mass_flux!(F,dΩ,V,RTMMchol,u₁*h₁)
```
amartinhuertas, line 51: To give the following a though (if u do not see it, forget it). 

Is it possible to recast `shallow_water_explicit_time_step!` such that  `thermal_shallow_water_explicit_time_step!` is able to reuse some code from the former? I see some common patterns among both solvers. It seems the thermal one is like some sort of extension to the regular one.

src/ThermalShallowWaterExplicit.jl
```diff
@@ -0,0 +1,219 @@
...
+      dump_diagnostics_thermal_shallow_water!(h_tmp, w_tmp,
+                                              model, dΩ, dω, S, L2MM, H1MM,
+                                              hn, un, En, wn, ϕ, F, eF, 1, dt,
+                                              diagnostics_file,
+                                              dump_diagnostics_on_screen)
+    end
+
+    # subsequent steps, do leap frog integration
+    # (now that we have the state at two previous time levels)
+    for istep in 2:N
```
amartinhuertas, line 176: I am starting to get a bit worried about the explosion that we have in time stepping loops. The logic within the loop is almost verbatim for all solvers. See also https://github.com/gridapapps/GridapGeosciences.jl/pull/22#issuecomment-932746770. Perhaps at some point we could start thinking how to abstract this into a common piece of code reusable by all solvers so that we mitigate this issue.

amartinhuertas: In any case, abstracting the thing into a common structure may come at loss of fliexibility to some extent if we do not design it with care/thinking about all possibilities. At this point, I am not sure if this is going to be possible/convenient. Writing my thoughs here ...

src/Helpers.jl
davelee2804: Good idea, I've modified the interface of the `new_vtk_step` as suggested

src/ThermalShallowWaterExplicit.jl
davelee2804: I've moved these into a separate function, `project_shallow_water_initial_conditions()`, and am calling this in the explicit, IMEX and thermal sovlers

src/ThermalShallowWaterExplicit.jl
davelee2804: This I don't know. Originally I wanted to pass in a function point to a method (explicit, IMEX, etc) to the time stepper, but this was inadequate as the different methods require different data, so I'm not sure how to generalise this...

src/ThermalShallowWaterExplicit.jl
davelee2804: Again, not sure. the Rosenbrock solver loops from 1:N not 2:N, and they all require different data, as mentioned above. With polymorphism I'm guessing it could be done (where each solver was a class with its own implementation of some virtual time_stepper and diagnostics, but doing this in a functional language I'm not sure...

Issue #13, opened 2021-08-30
Bottleneck in current cubed sphere mesh generation approach


amartinhuertas:
Highly related to https://github.com/gridap/Gridap.jl/issues/646 and https://github.com/gridap/Gridap.jl/issues/641

The current approach that we are using for generating the cubed sphere mesh has a bottleneck that will prevent us from scaling beyond hundreds of cells per panel (indeed, my laptop runs out of memory for approx 200x200 cells/panel). 

The issue is in this line:

https://github.com/gridap/Gridap.jl/blob/a52bc10503f77f700b7cee096c7311aebc131cee/src/Geometry/CartesianDiscreteModels.jl#L53

Recall that we generate a Cartesian volume mesh of the cube, which we then restrict to the cube surface. In this particular line, the Cartesian Grid is (temporarily) transformed into a UnstructuredGrid (coordinates vertices and connectivity explicitly stored, instead of managed implicitly), and this causes memory to increase cubically with the number of cells per direction (i.e., if we multiply the number of cells per direction by 2, the memory grows by a factor of 8). This is clearly not acceptable, it should increase quadratically, as we have a 2D mesh.

Possible approaches that comes into my mind (perhaps there are better solutions):
 * To bypass this issue is to write an ad-hoc extension of DiscreteModel{2,3} for the cubed sphere mesh made of 6 nxn CartesianDiscreteModel{2,2} panels joined together (as suggested by @davelee2804 in one of his comments) + the alpha/beta equiangular coordinates to R^3 coordinates that @cbladwell already tested previously.
 * Use p4est to handle the cubed sphere mesh. This would be required for large scale simulations.

davelee2804:
Hey @amartinhuertas , 

Looking to the short-medium term (ie: the current paper, any maybe a thermal shallow water equations paper), the test for this can be run on a mesh with 32X32 elements in each panel using RT2 (maybe also 48X48 elements using RT1), with a run time of 20 days and a 30 second time step for the explicit schemes and a 360 second time step for the implicit scheme.

It is still unclear to me if this resolution can be practically run using GridapPardiso with multi-threading and serial FEM assembly with the current mesh (maybe we might just get away with it :))

Your option 1 would probably get us over the line, if our current focus is to "do the paper"

However your option 2 is obviously the best practice approach. This could potentially form a paper in itself (the NUMA team have already published a paper on this topic, as you know). 

[...]

PR #25, opened 2021-11-16
Necessary mods to port repo to Gridap 0.17

amartinhuertas:
@davelee2804 @cbladwell @santiagobadia 

In this PR I have performed the necessary changes in GridapGeosciences s.t. it is now compatible with Gridap 0.17. Note that these changes **ONLY** affect to the internals of GridapGeosciences. The user drivers and tests have not been affected by these changes.

Besides, I have removed the hand-written computation of the Jacobian by an AD Jacobian in the Full Newton, implicit trapezoidal rule time stepper. I checked that this reduces the first call latency of  `Williamson2ThetaMethodFullNewtonTests.jl`  on my local machine with Julia 1.5.2 by approx 40% (1592.564583 secs. versus 932.869358 sec.)

@davelee2804 ... is it ok if we merge these changes into `master`? Does these changes affect in any way your current ongoing developments with GridapGeosciences?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [compat]
-Gridap = "0.17"
+Gridap = "0.17.2"
 SparseMatricesCSR = "0.6.2"
 julia = "1.5"
 
--- a/Project.toml
+++ b/Project.toml
@@ -20,9 +20,9 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [compat]
+Gridap = "0.17"
+SparseMatricesCSR = "0.6.2"
 julia = "1.5"
-SparseMatricesCSR ="0.6.2"
-Gridap = "0.16"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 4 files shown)


davelee2804:
Hi @amartinhuertas , thank you for taking the initiative on this. I am happy for you to merge these. 

I am currently doing some work on the thermal swes:
https://github.com/gridapapps/GridapGeosciences.jl/tree/tsw_thermogeostrophic
but I don't think this should conflict with your updates. I will pull your changes into my branch once they are merged.

amartinhuertas:
Ahhh ... I forgot to mention something!

This PR also solves https://github.com/gridapapps/GridapGeosciences.jl/issues/13 for the sequential cubed sphere mesh. This is a plus which comes with Gridap 0.17.

davelee2804:
Thanks @amartinhuertas . Is this using the P4est or custom 6n^2 mesh?

amartinhuertas:
> This PR also solves #13 for the sequential cubed sphere mesh. This is a plus which comes with Gridap 0.17.

After taking a much careful look at the new code, I have realized that this statement is false. Sorry about that! For the **sequential** `CubedSphereDiscreteModel` we still generate a volume cartesian mesh of the cube (with the current approach).
I was confused by another scenario in which Gridap 0.17 let us handle a manifold mesh without background volume mesh (e.g., when you read a surface mesh from data files). This does not mean that the issue cannot be solved with the current machinery in Gridap's Geometry module, but it requires some extra work. Anyway, I would not carry out this work because of the following.

> Is this using the P4est or custom 6n^2 mesh?

Using P4est, i.e., CubedSphere**Distributed**DiscreteModel, this issue is solved. It was solved even before Gridap 0.17. We only have this bottleneck for the sequential CubedSphereDiscreteModel. As I said above, it can be solved, but requires work that I would not do.

davelee2804:
No worries @amartinhuertas ,  as you say, probably not worth investing in the serial implementation too much. At high resolutions the serial code will have no shortage of performance problems :)

PR #26, opened 2021-11-17
GridapGeosciences Parallelization

amartinhuertas:
This is **working in progress**, thus I open the PR in draft mode so that we may start familarizing with the work I am doing, discuss whatever is needed, etc.

The final goal of this PR is introduce the MPI-parallel version of `GridapGeosciences.jl`. Some overarching goals are:
  * Maximize reuse of code among the sequential and parallel codes (this is possible as the high-level API of Gridap and GridapDistributed match).
  * Separate among sequential and parallel codes wherever appropiate (e.g., `CubedSphereDiscreteModel` versus p4est-based `CubedSphereDistributedDiscreteModel`). To this end, we try to separate the `src` and `test` source directories into `sequential` and `mpi` subdirectories.

This PR introduces a bunch of new major packages which are required to support distributed-memory parallelization, namely:
* `MPI.jl` - Julia wrappers for MPI.
* `PartitionedArrays.jl` - Julia native distributed linear algebra and communication layer.
* `GridapPETSc.jl` - Gridap/GridapDistributed wrappers for PETSc.jl written in Julia.
* `GridapP4est.jl` - Gridap wrappers for p4est written in Julia.
* `GridapDistributed.jl` - Parallel finite element library.

**[Solved]** Some of these are still under active development, with yet unreleased ongoing versions, so that we may have to point to these via the `Manifest.toml` file as usual.


src/mpi/CubedSphereDistributedDiscreteModels.jl
```diff
@@ -0,0 +1,349 @@
...
+          current=current+1
+         end
+         current_cell=current_cell+1
+       end
+     end
+     Gridap.Arrays.Table(data,ptr), panels
+  end
+end
+
+function generate_cube_grid_geo(cell_coordinates_and_panels)
```
davelee2804, line 133: Hi @amartinhuertas , Is this ordering of the panels consistently ordered? I think not right? (Perhaps I'm mistaken) We can consistently order the panels (so no sign_flip required) if we always travel in the same diagonal direction through the panels. Does that make sense?

amartinhuertas: Hi @davelee2804, as far I remember, for this particular data structure (parallel cubed sphere), I reused the same ordering of panels that was used in the UMKO cubed sphere meshes files that you shared. As a result, in here  see https://github.com/gridapapps/GridapGeosciences.jl/pull/26/files I do not have to explicitly change the sign of the normal to get it pointing outwards for some of the panels, as I had to do with the sequential data structure in here https://github.com/gridapapps/GridapGeosciences.jl/blob/master/src/CubedSphereDiscreteModels.jl#L157

Thus, answering your question, I think that in the parallel data structure the panels are consistently oriented with respect to the sphere manifold, but not necessarily oriented equivalently as in the sequential data structure.

davelee2804: Thanks for the clarification @amartinhuertas . If necessary we can always re-order, but great to know that it is already oriented.

amartinhuertas: See also https://github.com/gridapapps/GridapGeosciences.jl/pull/26/files#diff-ade293314c9bbc1cbab7c1aa0f90c560811028373d99f66a8a9bc7fa6212f8f3R26 and attached picture 
Screenshot from 2021-11-17 17-43-36

davelee2804: Thanks @amartinhuertas , I can see that if the two left panels are rotated pi/2 clockwise and the top panel is rotated pi/2 anti-clockwise the mesh is consistently oriented. Thanks for sharing.

amartinhuertas:
> Some of these are still under active development, with yet unreleased ongoing versions, so that we may have to point to these via the Manifest.toml file as usual.

This has been fixed in 02630c7. Let us see if the tests pass.

amartinhuertas:
Ok, tests are passing now! :muscle:  Accepting PR !

PR #27, opened 2021-11-18
get_normal_vector is now a method of triangulation

amartinhuertas:
In order to be consistent with Gridap (e.g., `get_facet_normal` is a method of `Triangulation`), and given that in Gridap 0.17 a triangulation is conceptually allowed to have access to the model, I have rewritten `get_normal_vector` such that it is now a method of the `CubedSphereTriangulation`s.

### DIFF
--- a/src/CubedSphereDiscreteModels.jl
+++ b/src/CubedSphereDiscreteModels.jl
@@ -45,6 +45,16 @@ function generate_Γface_to_bgface(model)
   Γface_to_bgface_panelwise
 end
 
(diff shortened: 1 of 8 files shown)


amartinhuertas:
Hi @davelee2804 ... are you happy that I accept this PR? It is a minor but important change towards parallelization using GridapDistributed

davelee2804:
Hi @amartinhuertas , yes, I saw that change. Happy for you to accept. It is an easy update to manage

PR #32, opened 2021-12-02
Thermal mat adv

davelee2804:
This involves the following changes:

1. A new solver for the thermal shallow water equations for which the buoyancy advection is in material form, with the buoyancy in the H1 conforming space (not L2). This seems to be more robust that the existing solver for the two tests that have been performed
2. A new "thermogeostrophic" configuration for the thermal shallow water equations. Basically a geostrophically balanced initial state with a thermally balanced state on top of that, and then a gaussian perturbation to the depth field in order to trigger some non-linearity around that state
3. A bug fix for the existing thermal SW diagnostics

In the next PR the upwinded trial funcs will be added, which will allow for additional high order energy consistent stabilisation of the potential vorticty and buoyancy fields.

### DIFF
--- a/driver/GalewskyThermalShallowWaterExplicit.jl
+++ b/driver/GalewskyThermalShallowWaterExplicit.jl
@@ -20,6 +20,16 @@ function gh₀(xyz)
   E
 end
 
+function g₀(xyz)
+  θϕr   = xyz2θϕr(xyz)
+  x,y,z = xyz
+  θ,ϕ,r = θϕr
+  α     = 1.0/3.0
+  β     = 1.0/10.0 #1.0/15.0
+  ϕ₂    = π/4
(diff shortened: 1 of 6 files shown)


amartinhuertas:
Thanks for your work @davelee2804 ! I have accepted the PR. The only thing I am a little bit concern is that it seems we are somehow repeating a bunch of code when introducing new variants of the equations/solvers. I am concerned that we could end up with an explosion of code replication. Perhaps there is a way to factor further the code into common functionality. Let us think a little bit about this in the meantime.

davelee2804:
Thanks @amartinhuertas , You make a good point. I requested this PR as without upwinding of either the facet integrals or the basis functions this was the only formulation in which I could effectively upwind the buoyancy. Once we have a better method of upwinding implemented maybe we can consolidate as you suggest....

PR #34, opened 2022-06-24
HDG for advection-diffusion on the sphere

amartinhuertas:
Super WIP. Opening PR in draft mode ... just for discussion purposes

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -148,10 +148,9 @@ deps = ["Artifacts", "Libdl"]
 uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
 
 [[Contour]]
-deps = ["StaticArrays"]
-git-tree-sha1 = "9f02045d934dc030edad45944ea80dbd1f0ebea7"
+git-tree-sha1 = "a599cfb8b1909b0f97c5e1b923ab92e1c0406076"
 uuid = "d38c429a-6771-53c6-b99e-75d170b6e991"
-version = "0.5.7"
+version = "0.6.1"
 
 [[Crayons]]
 git-tree-sha1 = "249fe38abf76d48563e2f4556bebd215aa317e15"
@@ -402,7 +401,7 @@ version = "0.2.4"
 
 [[GridapHybrid]]
 deps = ["DataFrames", "DrWatson", "FillArrays", "Gridap", "LinearAlgebra", "Plots", "ProfileView", "StaticArrays", "Test"]
-git-tree-sha1 = "d14b3d6e6cecb711309e040d071dcef97fa99c49"
+git-tree-sha1 = "7dfd95cfd5dd3a214978aed8cad1ffbfc8fe1bf1"
 repo-rev = "main"
 repo-url = "https://github.com/gridap/GridapHybrid.jl"
 uuid = "94e6dd8f-308f-4a47-89a7-811b3127ae57"
@@ -668,7 +667,7 @@ uuid = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
 version = "7.1.1"
 
 [[LinearAlgebra]]
-deps = ["Libdl", "libblastrampoline_jll"]
+deps = ["Libdl"]
 uuid = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [[LogExpFunctions]]
@@ -709,10 +708,10 @@ deps = ["Base64"]
 uuid = "d6f4376e-aef5-505a-96c1-9c027394607a"
 
 [[MbedTLS]]
-deps = ["Dates", "MbedTLS_jll", "Random", "Sockets"]
-git-tree-sha1 = "1c38e51c3d08ef2278062ebceade0e46cefc96fe"
+deps = ["Dates", "MbedTLS_jll", "MozillaCACerts_jll", "Random", "Sockets"]
+git-tree-sha1 = "891d3b4e8f8415f53108b4918d0183e61e18015b"
 uuid = "739be429-bea8-5141-9913-cc70e7f3736d"
-version = "1.0.3"
+version = "1.1.0"
 
 [[MbedTLS_jll]]
 deps = ["Artifacts", "Libdl"]
@@ -786,13 +785,9 @@ version = "1.3.5+1"
 
 [[OpenBLAS32_jll]]
 deps = ["Artifacts", "CompilerSupportLibraries_jll", "JLLWrappers", "Libdl", "Pkg"]
-git-tree-sha1 = "9c6c2ed4b7acd2137b878eb96c68e63b76199d0f"
+git-tree-sha1 = "ba4a8f683303c9082e84afba96f25af3c7fb2436"
 uuid = "656ef2d0-ae68-5445-9ca0-591084a874a2"
-version = "0.3.17+0"
-
-[[OpenBLAS_jll]]
-deps = ["Artifacts", "CompilerSupportLibraries_jll", "Libdl"]
-uuid = "4536629a-c528-5b80-bd46-f80d51c5b363"
+version = "0.3.12+1"
 
 [[OpenLibm_jll]]
 deps = ["Artifacts", "Libdl"]
@@ -852,10 +847,10 @@ uuid = "8fa3689e-f0b9-5420-9873-adf6ccf46f2d"
 version = "3.16.5+5"
 
 [[PackageCompiler]]
-deps = ["Artifacts", "LazyArtifacts", "Libdl", "Pkg", "Printf", "RelocatableFolders", "UUIDs"]
-git-tree-sha1 = "b2a8aef45444677037af79f060f5cb2ed48a72c4"
+deps = ["Artifacts", "LazyArtifacts", "Libdl", "Pkg", "Printf", "RelocatableFolders", "TOML", "UUIDs"]
+git-tree-sha1 = "52ab501c2201e140954924365ef06a36ba1d97ed"
 uuid = "9b87118b-4619-50d2-8e1e-99f35a4d4d9d"
-version = "2.0.6"
+version = "2.0.7"
 
 [[Pango_jll]]
 deps = ["Artifacts", "Cairo_jll", "Fontconfig_jll", "FreeType2_jll", "FriBidi_jll", "Glib_jll", "HarfBuzz_jll", "JLLWrappers", "Libdl", "Pkg"]
@@ -958,7 +953,7 @@ deps = ["InteractiveUtils", "Markdown", "Sockets", "Unicode"]
 uuid = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 
 [[Random]]
-deps = ["SHA", "Serialization"]
+deps = ["Serialization"]
 uuid = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [[RecipesBase]]
@@ -1384,10 +1379,6 @@ deps = ["Artifacts", "Bzip2_jll", "FreeType2_jll", "FriBidi_jll", "HarfBuzz_jll"
 git-tree-sha1 = "5982a94fcba20f02f42ace44b9894ee2b140fe47"
 uuid = "0ac62f75-1d6f-5e53-bd7c-93b484bb37c0"
 version = "0.15.1+0"
-
-[[libblastrampoline_jll]]
-deps = ["Artifacts", "Libdl", "OpenBLAS_jll"]
-uuid = "8e850b90-86db-534c-a0d3-1478176c7d93"
 
 [[libfdk_aac_jll]]
 deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,1421 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[ATK_jll]]
+deps = ["Artifacts", "Glib_jll", "JLLWrappers", "Libdl", "Pkg"]
+git-tree-sha1 = "58c36d8a1beeb12d63921bcfaa674baf30a1140e"
+uuid = "7b86fcea-f67b-53e1-809c-8f1719c154e8"
+version = "2.36.1+0"
+
+[[AbstractTrees]]
+git-tree-sha1 = "03e0550477d86222521d254b741d470ba17ea0b5"
+uuid = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
+version = "0.3.4"
+
+[[Adapt]]
+deps = ["LinearAlgebra"]
+git-tree-sha1 = "af92965fb30777147966f58acb05da51c5616b5f"
+uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+version = "3.3.3"
+
+[[ArgParse]]
+deps = ["Logging", "TextWrap"]
+git-tree-sha1 = "3102bce13da501c9104df33549f511cd25264d7d"
+uuid = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
+version = "1.1.4"
+
+[[ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+
+[[ArrayInterfaceCore]]
+deps = ["LinearAlgebra", "SparseArrays", "SuiteSparse"]
+git-tree-sha1 = "5e732808bcf7bbf730e810a9eaafc52705b38bb5"
+uuid = "30b0a656-2188-435a-8636-2ec0e6a096e2"
+version = "0.1.13"
+
+[[ArrayLayouts]]
+deps = ["FillArrays", "LinearAlgebra", "SparseArrays"]
+git-tree-sha1 = "56c347caf09ad8acb3e261fe75f8e09652b7b05b"
+uuid = "4c555306-a7a7-4459-81d9-ec55ddd5c99a"
+version = "0.7.10"
+
+[[Artifacts]]
+uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
+
+[[BSON]]
+git-tree-sha1 = "306bb5574b0c1c56d7e1207581516c557d105cad"
+uuid = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+version = "0.3.5"
+
+[[Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[BlockArrays]]
+deps = ["ArrayLayouts", "FillArrays", "LinearAlgebra"]
+git-tree-sha1 = "21490270d1fcf2efa9ddb2126d6958e9b72a4db0"
+uuid = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
+version = "0.16.11"
+
+[[Bzip2_jll]]
+deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
+git-tree-sha1 = "19a35467a82e236ff51bc17a3a44b69ef35185a2"
+uuid = "6e34b625-4abd-537c-b88f-471c36dfa7a0"
+version = "1.0.8+0"
+
+[[CEnum]]
+git-tree-sha1 = "eb4cb44a499229b3b8426dcfb5dd85333951ff90"
+uuid = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+version = "0.4.2"
+
+[[CSV]]
+deps = ["CodecZlib", "Dates", "FilePathsBase", "InlineStrings", "Mmap", "Parsers", "PooledArrays", "SentinelArrays", "Tables", "Unicode", "WeakRefStrings"]
+git-tree-sha1 = "873fb188a4b9d76549b81465b1f75c82aaf59238"
+uuid = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
+version = "0.10.4"
+
+[[Cairo]]
+deps = ["Cairo_jll", "Colors", "Glib_jll", "Graphics", "Libdl", "Pango_jll"]
+git-tree-sha1 = "d0b3f8b4ad16cb0a2988c6788646a5e6a17b6b1b"
+uuid = "159f3aea-2a34-519c-b102-8c37f9878175"
+version = "1.0.5"
+
+[[Cairo_jll]]
+deps = ["Artifacts", "Bzip2_jll", "Fontconfig_jll", "FreeType2_jll", "Glib_jll", "JLLWrappers", "LZO_jll", "Libdl", "Pixman_jll", "Pkg", "Xorg_libXext_jll", "Xorg_libXrender_jll", "Zlib_jll", "libpng_jll"]
+git-tree-sha1 = "4b859a208b2397a7a623a03449e4636bdb17bcf2"
+uuid = "83423d85-b0ee-5818-9007-b63ccbeb887a"
+version = "1.16.1+1"
+
+[[ChainRulesCore]]
+deps = ["Compat", "LinearAlgebra", "SparseArrays"]
+git-tree-sha1 = "9489214b993cd42d17f44c36e359bf6a7c919abf"
+uuid = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+version = "1.15.0"
+
+[[ChangesOfVariables]]
+deps = ["ChainRulesCore", "LinearAlgebra", "Test"]
+git-tree-sha1 = "1e315e3f4b0b7ce40feded39c73049692126cf53"
+uuid = "9e997f8a-9a97-42d5-a9f1-ce6bfc15e2c0"
+version = "0.1.3"
+
+[[CodecZlib]]
+deps = ["TranscodingStreams", "Zlib_jll"]
+git-tree-sha1 = "ded953804d019afa9a3f98981d99b33e3db7b6da"
+uuid = "944b1d66-785c-5afd-91f1-9de20f533193"
+version = "0.7.0"
+
+[[ColorSchemes]]
+deps = ["ColorTypes", "ColorVectorSpace", "Colors", "FixedPointNumbers", "Random"]
+git-tree-sha1 = "1fd869cc3875b57347f7027521f561cf46d1fcd8"
+uuid = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
+version = "3.19.0"
+
+[[ColorTypes]]
+deps = ["FixedPointNumbers", "Random"]
+git-tree-sha1 = "eb7f0f8307f71fac7c606984ea5fb2817275d6e4"
+uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
+version = "0.11.4"
+
+[[ColorVectorSpace]]
+deps = ["ColorTypes", "FixedPointNumbers", "LinearAlgebra", "SpecialFunctions", "Statistics", "TensorCore"]
+git-tree-sha1 = "d08c20eef1f2cbc6e60fd3612ac4340b89fea322"
+uuid = "c3611d14-8923-5661-9e6a-0046d554d3a4"
+version = "0.9.9"
+
+[[Colors]]
+deps = ["ColorTypes", "FixedPointNumbers", "Reexport"]
+git-tree-sha1 = "417b0ed7b8b838aa6ca0a87aadf1bb9eb111ce40"
+uuid = "5ae59095-9a9b-59fe-a467-6f913c188581"
+version = "0.12.8"
+
+[[Combinatorics]]
+git-tree-sha1 = "08c8b6831dc00bfea825826be0bc8336fc369860"
+uuid = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
+version = "1.0.2"
+
+[[CommonSubexpressions]]
+deps = ["MacroTools", "Test"]
+git-tree-sha1 = "7b8a93dba8af7e3b42fecabf646260105ac373f7"
+uuid = "bbf7d656-a473-5ed7-a52c-81e309532950"
+version = "0.3.0"
+
+[[Compat]]
+deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "SHA", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
+git-tree-sha1 = "9be8be1d8a6f44b96482c8af52238ea7987da3e3"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "3.45.0"
+
+[[CompilerSupportLibraries_jll]]
+deps = ["Artifacts", "Libdl"]
+uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
+
+[[Contour]]
+deps = ["StaticArrays"]
+git-tree-sha1 = "9f02045d934dc030edad45944ea80dbd1f0ebea7"
+uuid = "d38c429a-6771-53c6-b99e-75d170b6e991"
+version = "0.5.7"
+
+[[Crayons]]
+git-tree-sha1 = "249fe38abf76d48563e2f4556bebd215aa317e15"
+uuid = "a8cc5b0e-0ffa-5ad4-8c14-923d3ee1735f"
+version = "4.1.1"
+
+[[DataAPI]]
+git-tree-sha1 = "fb5f5316dd3fd4c5e7c30a24d50643b73e37cd40"
+uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
+version = "1.10.0"
+
+[[DataFrames]]
+deps = ["Compat", "DataAPI", "Future", "InvertedIndices", "IteratorInterfaceExtensions", "LinearAlgebra", "Markdown", "Missings", "PooledArrays", "PrettyTables", "Printf", "REPL", "Reexport", "SortingAlgorithms", "Statistics", "TableTraits", "Tables", "Unicode"]
+git-tree-sha1 = "daa21eb85147f72e41f6352a57fccea377e310a9"
+uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
+version = "1.3.4"
+
+[[DataStructures]]
+deps = ["Compat", "InteractiveUtils", "OrderedCollections"]
+git-tree-sha1 = "d1fff3a548102f48987a52a2e0d114fa97d730f0"
+uuid = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
+version = "0.18.13"
+
+[[DataValueInterfaces]]
+git-tree-sha1 = "bfc1187b79289637fa0ef6d4436ebdfe6905cbd6"
+uuid = "e2d170a0-9d28-54be-80f0-106bbe20a464"
+version = "1.0.0"
+
+[[Dates]]
+deps = ["Printf"]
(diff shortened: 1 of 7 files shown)


davelee2804:
Hey @amartinhuertas , I'm having some basic trouble with the HDG advection on the sphere. I am getting the error:
> It is not possible to perform the operation "dot" on the given cell fields.

here:
https://github.com/gridapapps/GridapGeosciences.jl/blob/hdg_adv/src/AdvectionHDG.jl#L19
Currently I am passing in `un` directly from the function:
https://github.com/gridapapps/GridapGeosciences.jl/blob/hdg_adv/driver/sequential/SBRAdvectionHDG.jl#L19
However I have also tried projecting `un` onto a FESpace via
https://github.com/gridapapps/GridapGeosciences.jl/blob/hdg_adv/src/AdvectionHDG.jl#L57
https://github.com/gridapapps/GridapGeosciences.jl/blob/hdg_adv/src/AdvectionHDG.jl#L90
I'm wondering if the vector field on the manifold and the L2 reference element vector field (as I've defined it) are somehow incompatible? (Note that D = 2)...

amartinhuertas:
Can you please separate the terms in the bilinear form one by one as in e.g., https://github.com/gridap/GridapHybrid.jl/blob/main/test/DarcyHDGTests.jl#L71
and let me know which is the one(s) that triggers the error? Thanks!

amartinhuertas:
> (Note that D = 2).

Did you try with D=3?

davelee2804:
...I tried `D=3`, but it failed even earlier (I can't remember where exactly sorry). Thanks for the suggestion, the failure is at:
`  _op2 = (un⋅n)`

amartinhuertas:
can you attach a file with the whole backtrace? I would like to understand the cause behind the inability to perform the dot product operator.

davelee2804:
[err.log](https://github.com/gridapapps/GridapGeosciences.jl/files/9016173/err.log)

amartinhuertas:
Ok, digging into the error log, I saw this line of code:

> setindex!(A::Vector{Gridap.TensorValues.VectorValue{2, Float64}}, x::Gridap.TensorValues.VectorValue{3, Float64}, i1::Int64)

There we are trying to set a 3-components field vector (`Gridap.TensorValues.VectorValue{3, Float64}`) as an entry of a Vector of 2-component field vectors (`Vector{Gridap.TensorValues.VectorValue{2, Float64}}`). Thus, as you say, it seems we are messing things up with the number of spatial dimensions. 

What is the error that you get with D=3?

davelee2804:
Here is the D=3 log file:
[err-2.log](https://github.com/gridapapps/GridapGeosciences.jl/files/9016220/err-2.log)

amartinhuertas:
ok. Use D=3 ONLY for this line, i.e., 

`reffeᵤ = ReferenceFE(lagrangian,VectorValue{3,Float64},order;space=:P)`

The rest untouched.

davelee2804:
...Same error as the original, as far as I can see: :( 
[err-3.log](https://github.com/gridapapps/GridapGeosciences.jl/files/9016283/err-3.log)

davelee2804:
...I just got this error:
[err-4.log](https://github.com/gridapapps/GridapGeosciences.jl/files/9016349/err-4.log)
...with this source:
[AdvectionHDG.log](https://github.com/gridapapps/GridapGeosciences.jl/files/9016350/AdvectionHDG.log)
(so `un` is here projected onto `U`, but is maybe not being interpreted as a vector field)?

amartinhuertas:
> Hey @amartinhuertas , I'm having some basic trouble with the HDG advection on the sphere. I am getting the error:

For the records, this was addressed in https://github.com/gridapapps/GridapGeosciences.jl/pull/34/commits/1586cf7729879573c9598fe9dcf56acfad16db25

amartinhuertas:
Hi @davelee2804, 

please read the below carefully, let me know if you have questions, and please answer my questions (no hurries).

Current status of this development:
* I have solved several minor issues in `GridapHybrid.jl` when dealing with manifolds. (see https://github.com/gridap/GridapHybrid.jl/pull/23/files). These have been already merged to `GridapHybrid.jl#main`. As we are pointing there in the `GridapGeosciences.jl#hdg_adv` branch, after pulling from remote `hdg_adv`, you just have to write `up` in the Julia REPL package manager to update `GridapHybrid.jl`.
* Now the HDG driver for advection on the sphere runs bottom-top (it does not crash). **HOWEVER** it does not produce the correct result, in particular, it does not seem to time evolve. 

Important remarks:
* Let us stick (by now) to a 2nd-order polynomial approximation of the sphere (note that in the latest version of the driver I am passing polynomial degree `2` in this call https://github.com/gridapapps/GridapGeosciences.jl/pull/34/files#diff-d6ace757a93640319c464479d34d4617d5d53c419c630a9afee6b9c8db9b8d46R41 ). I still have to go through the analytical mapping cubed sphere mesh to see whether it properly represents the 1D skeleton in 3D ambient space. 
* I had to change the order of the operands in terms like this `m*p*((un⋅nₒ) + τ*(n⋅nₒ))` as `((un⋅nₒ) + τ*(n⋅nₒ))*m*p`. In the former form, `GridapHybrid.jl` generates an internal error (I will solve it when I have time, I think there are some functions missing, but I still have to fully understand this). However, in the latter form, it does not.

Questions:
* Do you have a reference where the formulation you are aiming to implement is written? If not, can you write it? This may help to (1) try to understand what you are trying to do. (2) try to determine if the code is doing what you are trying to do or not.
* I have seen you have terms like `n \cdot n`. Given that n is a unit vector, isnt `n \cdot n` always 1?
* Note that `no` and `n` are defined differently. `n` is the tangent outward unit normal (it always points outwards). `no` points outwards if the corresponding cell is the OWNER of the edge, and points inwards if the corresponding cell is not the owner. Thus, for a given cell K, and edge E in the bounday of K, `n \cdot no` is 1 if K owns E, and it is not 1 (its value depends on the cosine of the angle among the two normals) if it does not own it. We assign the owner cell of an edge arbitrarily to be the cell with the minimum global identifier among the two cells surrounding the edge.

davelee2804:
Thank you @amartinhuertas , that was quick!

I'm not surprised there are still bugs in the code. I update to your current version of GridapHybrid.jl and take a look (hopefully tomorrow).

> Let us stick (by now) to a 2nd-order polynomial approximation of the sphere

Great idea, yes, let's do this

> Do you have a reference where the formulation you are aiming to implement is written?

It is solid body rotation of the flux form advection equation, so the (constant) velocity field is 0 in the north-south direction and `cos(\phi)` in the east-west direction (with `\phi` being the latitude). This is the 'Williamson 1' test case ('Advection of a Cosine Bell) from this paper here: https://www.sciencedirect.com/science/article/pii/S0021999105800166 (with `\alpha=0`). However I am using a Gaussian not a cosine as this is infinitely differentiable, so a little nicer to work with.

> Given that n is a unit vector, isnt n \cdot n always 1

Yes, as you say. I have kept the `n\cdot n` as I wanted to be consistent with respect to the `n\cdot n_o` terms in the jumps, just until I am comfortable with the method, then I will remove

> Note that no and n are defined differently

Yes, I noticed this, thanks. So `n\cdot n_o` represents a jump, ie a `[[ . ]]` operator yes?

amartinhuertas:
> I'm not surprised there are still bugs in HGD advection the code. I update to your current version of GridapHybrid.jl and take a look (hopefully tomorrow).

Please note that the bugs might not come only from your side. We are testing GridapHybrid with manifolds for the first time and there might be still pitfalls/bugs to be determined/fixed. I think you are aware of this, but just to be 100% sure. I would be extremely careful, and may be try to think in simpler tests we can write to determine whether integration of fields over \partialK is actually working as expected with manifolds.

 > The weak HDG form of the flux form advection equation is based on the formulation in Kang et. al. JCP 2020

Can you send me the particular Eq. number in that paper?  

> Yes, I noticed this, thanks. So n\cdot n_o represents a jump, ie a [[ . ]] operator yes?

Please note that, in general, the jump operator may come in two different flavours:
* `u^+ - u^-`, where u can be a scalar or vector double-valued discontinuous field. If u is scalar, then the jump is scalar, if it is a vector, then it is a vector. This kind of jumps are used in the 
* `u^+ n^+ + u^- n^-`. Where n is the outward unit normal as seen from the perspective of the two cells around.  If u is an scalar, the jump is a vector. If u is a vector, the jump is an scalar. Note that for manifolds n^+ is not equal to -n^- in general.

See def 14.4 and remark 14.5 in 
https://www.math.tamu.edu/~guermond/M610_SPRING_2016/chap14.pdf

n \cdot n0 seems actually another thing (like trying to give a global orientation to the normal, and then adjusting if  the normal from the current cell does not match the one given in the global orientation), but perhaps I am missing something.  I am not saying it is incorrect, I am bit confused actually.

davelee2804:
Hey @amartinhuertas , the relevant equations in Kang et. al. are (3a), (16) and (17).

With respect to the normals we also have that `n^+ = -n^-`, so assuming `K^+` is the cell owning the normal, then `n^+ = n_o` and `n^- = -n_o` so `u\cdot n\_o` integrated over `\partial K` from both sides would be `(n^+ - n^-)\cdot n_o` (ie: a scalar jump). At least that is my understanding. I also get very confused with these binary choices (that is why I left the `n\cdot n` terms in the formulation :) ), there is a very strong chance that I am wrong...

davelee2804:
Hey @amartinhuertas , just a heads up, when I check the `sum()` of the input fields `un`, `pn`, these are non-zero, however the half step field, `ph` has zero `sum()`. I've also looked at individual matrix terms as vectors (assuming pn for the trial function), and these also have non-zero `sum()`

amartinhuertas:
> u\cdot n\_o integrated over \partial K from both sides would be (n^+ - n^-)\cdot n_o

Why u\cdot n\_o integrated over both sides and then summed is (n^+ - n^-)\cdot n_o? In the latter expression, I do not see the u. ??? What am I missing ...

> With respect to the normals we also have that n^+ = -n^-

I have doubts in regards to this equivalence for discretized manifolds. I think this equivalence is only approximate in this case, as the normals might be contained in different planes (the ones defined by the cells). At least the code does not assume this, and compute n^+ and n^- from each cell without relying on this equivalence. I guess that the finer the mesh, the more accurate the approximation, but not totally sure. I can write an small numerical experiment to test whether this is true or not.

amartinhuertas:
> Hey @amartinhuertas , just a heads up, when I check the sum() of the input fields un, pn, these are non-zero, however the half step field, ph has zero sum(). I've also looked at individual matrix terms as vectors (assuming pn for the trial function), and these also have non-zero sum()

Thanks for reporting this! I found another BUG for GridapHybrid.jl + manifolds. I have fixed it in this PR https://github.com/gridap/GridapHybrid.jl/pull/24/files. I have merged it into `main`.  Please execute `up` in the package manager to get this BUG fix on your env. 

I have executed the HDG advection driver after the bug fix, and it now time evolves, but it does seem to generate just rubbish. We still have to find extra bug fixes, formulation pitfalls etc.

amartinhuertas:
@davelee2804  ... for convenience, I wrote a MWE (Minimum Working Example) to find the previous BUG. I am copying it here so that you can also leverage it if require. I think it can be highly helpful as a template to write MWEs for other yet-to-spot BUGs. 

```julia
using LinearAlgebra
using SparseArrays
using Gridap
using GridapGeosciences
using GridapHybrid


function Gridap.Geometry.push_normal(invJt,n)
  v = invJt⋅n
  m = sqrt(inner(v,v))
  if m < eps()
    return zero(v)
  else
    return v/m
  end
end

function project_initial_conditions(dΩ, P, Q, p₀, U, V, u₀, mass_matrix_solver)
  # the tracer
  b₁(q)    = ∫(q*p₀)dΩ
  a₁(p,q)  = ∫(q*p)dΩ
  rhs₁     = assemble_vector(b₁, Q)
  L2MM     = assemble_matrix(a₁, P, Q)
  L2MMchol = numerical_setup(symbolic_setup(mass_matrix_solver,L2MM),L2MM)
  pn       = FEFunction(Q, copy(rhs₁))
  pnv      = get_free_dof_values(pn)

  solve!(pnv, L2MMchol, pnv)

  # the velocity field
  b₂(v)    = ∫(v⋅u₀)dΩ
  a₂(u,v)  = ∫(v⋅u)dΩ
  rhs₂     = assemble_vector(b₂, V)
  MM       = assemble_matrix(a₂, U, V)
  MMchol   = numerical_setup(symbolic_setup(mass_matrix_solver,MM),MM)
  un       = FEFunction(V, copy(rhs₂))
  unv      = get_free_dof_values(un)

  solve!(unv, MMchol, unv)

  pn, pnv, L2MM, un
end

order  = 1
degree = 3
n      = 16
dt     = 0.25*2.0*π/((order+1)*4*n)
# single rotation about the sphere
nstep  = Int(2.0*π/dt)

# solid body rotation velocity field
function u₀(xyz)
  θϕr = xyz2θϕr(xyz)
  u   = cos(θϕr[2])
  spherical_to_cartesian_matrix(θϕr)⋅VectorValue(u,0,0)
end

# Gaussian tracer initial condition
function p₀(xyz)
  rsq = (xyz[1] - 1.0)*(xyz[1] - 1.0) + xyz[2]*xyz[2] + xyz[3]*xyz[3]
  exp(-4.0*rsq)
end

model = CubedSphereDiscreteModel(1,2; radius=1)

# Forward integration of the advection equation
D = num_cell_dims(model)
Ω = Triangulation(ReferenceFE{D},model)
Γ = Triangulation(ReferenceFE{D-1},model)
∂K = GridapHybrid.Skeleton(model)

reffeᵤ = ReferenceFE(lagrangian,VectorValue{3,Float64},order;space=:P)
reffeₚ = ReferenceFE(lagrangian,Float64,order;space=:P)
reffeₗ = ReferenceFE(lagrangian,Float64,order;space=:P)

# Define test FESpaces
V = TestFESpace(Ω, reffeᵤ; conformity=:L2)
Q = TestFESpace(Ω, reffeₚ; conformity=:L2)
M = TestFESpace(Γ, reffeₗ; conformity=:L2)
Y = MultiFieldFESpace([Q,M])

U = TrialFESpace(V)
P = TrialFESpace(Q)
L = TrialFESpace(M)
X = MultiFieldFESpace([P,L])

dΩ  = Measure(Ω,degree)
d∂K = Measure(∂K,degree)

# HDG stabilisation parameter
τ = 1.0

# Project the initial conditions onto the trial spaces
pn, pnv, L2MM, un = project_initial_conditions(dΩ, P, Q, p₀, U, V, u₀, Gridap.Algebra.BackslashSolver())

dth = 0.5*dt

n  = get_cell_normal_vector(∂K)
nₒ = get_cell_owner_normal_vector(∂K)


# First stage
b₁((q,m)) = ∫(q*pn)dΩ - ∫(m*0.0)d∂K
a₁((p,l),(q,m)) = ∫(q*p - dt*(∇(q)⋅un)*p)dΩ + ∫(((un⋅n) + τ*(n⋅n))*dt*q*p)d∂K -   # [q,p] block
              ∫(dt*q*l*τ*(n⋅n))d∂K +                                              # [q,l] block
              ∫(((un⋅nₒ) + τ*(n⋅nₒ))*m*p)d∂K -                                    # [m,p] block
              ∫(m*l*τ*(n⋅nₒ))d∂K                                                  # [m,l] block


a11(p,q)=∫(q*p - dt*(∇(q)⋅un)*p)dΩ + ∫(((un⋅n) + τ*(n⋅n))*dt*q*p)d∂K
A11=assemble_matrix(a11,X[1],Y[1])
a12(l,q)=∫(-dt*q*l*τ*(n⋅n))d∂K
A12=assemble_matrix(a12,X[2],Y[1])
a21(p,m)=∫(((un⋅nₒ) + τ*(n⋅nₒ))*m*p)d∂K
A21=assemble_matrix(a21,X[1],Y[2])
a22(l,m)=∫(-1.0*m*l*τ*(n⋅nₒ))d∂K
A22=assemble_matrix(a22,X[2],Y[2])

b1(q)= ∫(q*pn)dΩ
b2(m)=∫(-m*0.0)d∂K

A=assemble_matrix(a₁,X,Y)

B1=assemble_vector(b1,Y[1])
B2=assemble_vector(b2,Y[2])

BΓ  = B2  - Matrix(A21)*(A11\B1)
BΓΓ = A22 - Matrix(A21)*inv(Matrix(A11))*Matrix(A12)

op₁  = HybridAffineFEOperator((u,v)->(a₁(u,v),b₁(v)), X, Y, [1], [2])
Xh   = solve(op₁)
ph,_ = Xh

op_BΓΓ = op₁.skeleton_op.op.matrix
op_BΓ = op₁.skeleton_op.op.vector

@assert norm(op_BΓΓ-BΓΓ) < 1.0e-12
norm(BΓ-op_BΓ) < 1.0e-12
```

davelee2804:
Hey @amartinhuertas 

> Why u\cdot n_o integrated over both sides and then summed is (n^+ - n^-)\cdot n_o? In the latter expression, I do not see the u. ??? What am I missing ...

Sorry, silly typo, that should read `(u^+ - u^-)\cdot n_o`, silly typo on my part

> I think this equivalence is only approximate in this case, as the normals might be contained in different planes

Great point, yes, I think as one refines the mesh `n^+` _converges to_ `-n^-`, but it will probably not be exact (the H(div) Piola map gives this exactly though yes?)

amartinhuertas:
hi @davelee2804 ... I am trying to make sense out of the HDG formulation that you have written, but I am struggling quite a bit (I cannot say whether it is correct or not, just that I dont understand). 

Looking at https://arxiv.org/abs/1605.03228, Eqs. (3.1)-(3.4), they briefly present the HDG upwind formulation for the (steady-state) advection equation (originally proposed in ref [5] https://users.oden.utexas.edu/~tanbui/PublishedPapers/phoga.pdf). Is this the formulation that are you trying to implement? 

For convenience, the correspondence among the article notation and the code notation is:
 * `\beta -> un`
 * `u -> p`
 * `v -> q`
 * `\hat{u} -> l`
 * `\mu -> m`

By now, I have sevaral concerns:
 1. The article mentions that `beta` is assumed to be a continuous function. However, `un` being used in the code is the projection of a continuous function into a discontinuous FE space. Correct? I would like to understand why `beta` has to be continuous.
 2. The term which is integrated over the bulk (i.e., K) in the article does not match the one in the code. 
 3. I do not know how to proof the equivalence among the terms in the code and the terms in the article over \partial K.

Perhaps you are trying to implement a different formulation ... not sure ....


Screenshot from 2022-07-05 09-50-21

davelee2804:
Hey @amartinhuertas , 

> Is this the formulation that are you trying to implement?

They are doing material form advection `\beta.\nabla u`, whereas I am doing flux form `\nabla\cdot(\beta u()` For a divergence free vector field, `\nabla\cdot\beta = 0`, the two are the same. The velocity field we are using `(\cos(\phi),0)` is divergence free in the continuous form, but for `\beta\in H1(\Omega` I would imagine there will be some numerical divergence, so they won't be exactly the same

> However, un being used in the code is the projection of a continuous function into a discontinuous FE space. Correct? I would like to understand why beta has to be continuous.

We could do `un` continuous or discrete. We are only solving a single equation `u -> p` so there is the velocity field `\beta->un` will always be prescribed. When we extend to the shallow water equations then we solve for `\beta->un`

> The term which is integrated over the bulk (i.e., K) in the article does not match the one in the code.

Indeed, that is a result of the marerial vs. flux form advection. 

I don't understand the differences in the mesh skeleton term w.r.t. Kang et al 2020. (ie the use of `|\bea\cdot n|`) - do you have any thoughts on this??

amartinhuertas:
> They are doing material form advection \beta.\nabla u, whereas I am doing flux form \nabla\cdot(\beta u) For a divergence free vector field, \nabla\cdot\beta = 0, the two are the same.

Ok, I now understand that if the fluid flow is incompressible, then both the term in the code and the one in the article are equivalent, as I have double checked in the following note. Agreed?

Screenshot from 2022-07-06 21-27-11

amartinhuertas:
> We could do un continuous or discrete. 

I would do it continuous (as far as possible). Is there any reason you need to interpolate it into a FE space? The only concern is that if we have a differential operator applied to the continuous solution we have to ensure that we apply the spherical variant of the differential operator, as we do, e.g., in the Darcy test with the laplacian applied to the analytical solution.

> We are only solving a single equation u -> p so there is the velocity field \beta->un will always be prescribed. When we extend to the shallow water equations then we solve for \beta->un

Yes, I was aware of this. \beta is input data to the advection problem.

amartinhuertas:
> I don't understand the differences in the mesh skeleton term w.r.t. Kang et al 2020. (ie the use of |\bea\cdot n|) - do you have any thoughts on this??

I think that Kang et al and the iHDG paper (the one from which I extracted the capture above) use the same abstract family of HDG formulations for hyperbolic problems of the general form

Screenshot from 2022-07-06 21-45-36

For the advection problem (3.1), we have that 

```
F_k(u)=\beta_k u 
C=0
```

Thus, `A_k` is a 3x3 tensor with entries defined as `[A_k]_kk=\beta_k` and 0 otherwise. 

Agreed? (please let me know if everything clear)

amartinhuertas:
The general form of the upwind HDG discretization for problem (2.1) reads as:

Screenshot from 2022-07-06 21-51-22

In the case of problem (3.1), A is diagonal matrix, with \beta_k n_k in the diagonal, for k=1..3. Thus, after computing the eigendecomposition of A, R is the identity, and S is equal to A, and |A| is |\beta \cdot n|, which is precisely what they use in (3.2) and (3.3). This choice of the upwind flux is the solution of a Riemman problem on the interface of the cells (TBH, I still do not fully understand this part, I need to devote more time to understand where the definition of the \hat{F} flux comes from and why it provides the desired convergence, stability, etc. properties.).

I guess that in Kang et. al. they are leveraging the eigenvalues of the A matrix that appears for the shallow water equations, and thus this does not apply verbatim to the (much simpler) advection equation.

How did you derive the formulation in the code for the advection equation from the one  of the shallow water equations in Kang et. al.? Could this be what is wrong?

davelee2804:
Hey @amartinhuertas , I think I agree with what you say. If I unroll the definition of `\hat{F}\cdot n` from eqn (3) in Muralikrishnan et al as used in section (4) of that paper I get:

`\hat{F}\cdot n = \beta\cdot n u + |\beta\cdot n|(u - \hat{u})`, as in eqn (19), where `\beta` is the trial eigenvalue for the advection problem

if I unroll (17) from Kang et al (which is what I was trying to do) I get (using the same nomenclature) the same (using `n\cdot n = 1`)

Based on this I think my current implementation is wrong in two ways:
1. I am not using an absolute value, and 
2. I am not using the correct `\tau` (I am just using `\tau=1` for convenience, based on the max value of the absolute value of `|\beta\cdot n|` for this simple problem.

Is there a way to apply an absolute value in Gridap? I know there is `mean()` and `jump()`, is there also an `abs()` ?

amartinhuertas:
> I know there is mean() and jump(), is there also an abs() ?

I think that is possible. AFAIK, the high level API supports an invokation to an arbitrary Julia function passing a `CellField` as an argument (\beta \cdot n is a `CellField`). Anyway, I have never tried that with `GridapHybrid.jl` ... let me know if that does not work.

amartinhuertas:
BTW, did you notice that in (2.2a) they are using exactly the same bulk term as the one you are using in your code? :-)

davelee2804:
> BTW, did you notice that in (2.2a) they are using exactly the same bulk term as the one you are using in your code? :-)

I haven't checked it experimentally, but for a divergence free `\beta` I think it should be OK. 

I've just made a change that leads to a much more coherent solution (and exact mass conservation), however, there is no `jump()` in there now as this leads to a code error as: `It is not possible to perform the operation "*" on the given cell fields.`

Do you know if the `jump()` operator is supported for integration with respect to test functions on the Skeleton using the measure `d\partial K` ?

amartinhuertas:
> Do you know if the jump() operator is supported for integration with respect to test functions on the Skeleton using the measure d\partial K ?

Yes. The `jump()` operator is not supported, on purpose. The measure d\partial K represents the boundary of the cells, on a cell by cell basis, such that each cell is independent of each other. The `jump()` operator can be used only with facet-wise triangulations that glue together the two cells around. I am not sure why you need the jump() operator explicity.

davelee2804:
OK, great to know, thanks for the confirmation @amartinhuertas . Well, how do we represent `[[.]]` then? I guess this is what `n\_o` is for... I'll keep trying to find a way to replicate the jump() with this. Right now the results look coherant and conservative, but I think over long times they may go oscillatory without some sort of jump penalisation...

amartinhuertas:
> Well, how do we represent [[.]] then?

Let us assume that we want to compute, e.g., 

`< [[u]], v>_e`, for all edges `e` in the mesh, where `[[u]] = u^+ - u^-` 

Then you have to write 

`< u*(n \cdot n_o), v>_d\partial K`, for all cells `K` in the mesh.

So I think that you already got that from what I see in your comments.

amartinhuertas:
> Right now the results look coherant and conservative, but I think over long times they may go oscillatory without some sort of jump penalisation...

This is great! I am surprised that we only needed a moderate amount of BUG fixes. :-)

amartinhuertas:
> I haven't checked it experimentally, but for a divergence free \beta I think it should be OK.

I am surprised they do not mention that \beta is assumed to be divergence free at any point in the iHDG paper ... Perhaps it is something well-known in hyperbolic fluid flow problems or something like that ...

davelee2804:
> This is great! I am surprised that we only needed a moderate amount of BUG fixes. :-)

Doesn't mean it is working correctly yet however :). Could be that it will still go oscillatory or not advect correctly (I only checked a couple of time steps). 

When I add in the jumps as `n\cdot n\_o` I get completely nonsense results (and a total loss of mass conservation), so I am wondering if there is some subtle interplay between the `n\cdot n_o` jump and the `|A|` absolute value that I don't properly understand....

davelee2804:
> I am surprised they do not mention that \beta is assumed to be divergence free at any point in the iHDG paper ... Perhaps it is something well-known in hyperbolic fluid flow problems or something like that ...

`\beta` could have a divergence. It is just that if `\beta` has non-zero divergence then the material and flux form advection equations are no longer equivalent. I thought I would do flux form as that is how the advection equation is done in shallow water.

amartinhuertas:
> \beta could have a divergence. It is just that if \beta has non-zero divergence then the material and flux form advection equations are no longer equivalent. I thought I would do flux form as that is how the advection equation is done in shallow water.

Ok. Thanks for the clarification. I think I 100% understand now.

amartinhuertas:
> Doesn't mean it is working correctly yet however :). Could be that it will still go oscillatory or not advect correctly (I only checked a couple of time steps).
> When I add in the jumps as n\cdot n\_o I get completely nonsense results (and a total loss of mass conservation), so I am wondering if there is some subtle interplay between the n\cdot n_o jump and the |A| absolute value that I don't properly understand..

To which kind of oscillations are you referring to? You mean spurious modes that you usually have in the Galerkin method if you do not use upwinding? (such as SUPG?). 

I am not wrong, the formulation we are leveraging already comes with upwinding "batteries-on", no need to do anything else. But perhaps I am missing something ...

davelee2804:
Yes, the formulation we are trying to implement includes upwinding, agreed. However I am not able to get the jump penalisation working, which is why I think it will eventually go oscillatory...

davelee2804:
...Does the integral over the measure `d\partial K` already assume that the jump is baked in?

amartinhuertas:
> I am not able to get the jump penalisation working

Which is the evidence that supports this statement?

amartinhuertas:
> ...Does the integral over the measure d\partial K already assume that the jump is baked in?

No, AFAIK.

amartinhuertas:
Have you noticed that both Kang et al and MURALIKRISHNAN et al defined the jump as [[u]] = u^+ + u^- ??? With a + not a -

amartinhuertas:
> Which is the evidence that supports this statement?

Ok, I think you already answered to that, sorry. Namely here

> ie: the n\cdot n\_o term leads to large errors...

I am not sure we have to use n\cdot n\_o here, given the definition of the jump operator in that papers.

davelee2804:
Hey @amartinhuertas , as mentioned, the application of 'n\cdot n_o` leads to erroneous results for me. With the current configuration the results are correct (conservative and advecting at the correct speed), see images attached. So I'm very confused about the application of the jumps. In this article (which also discusses continuity of the tangent vectors), there are no jumps explicitly mentioned (see eqn 7 for example): https://www.sciencedirect.com/science/article/pii/S0021999111003226



time_0T
time_1T
time_Ton2
time_Ton4
:

davelee2804:
...so I have no idea if the penalisation is properly applied or not. Given that the results are non-oscillatory and conservative, it looks like _maybe_ this is being done correctly, but I don't know enough about how this works under the hood to say. Also I will need to implement a more sophisticated version of `\tau` (as `|u\cdot n|`, not `1`).

amartinhuertas:
> In this article (which also discusses continuity of the tangent vectors), there are no jumps explicitly mentioned (see eqn 7 for example): https://www.sciencedirect.com/science/article/pii/S0021999111003226

Hi @davelee2804 ! The jumps are defined here, yes? (Again, as in the other papers, with a + not a -, i.e., no need for n \cdot no)

Screenshot from 2022-07-08 17-45-09

amartinhuertas:
> Also I will need to implement a more sophisticated version of \tau (as |u\cdot n|, not 1).

In my experience, if \tau is not set to the appropriate value, one of the effects might be that the method does not convergence at the proper order as you refine the mesh.

amartinhuertas:
Note that an integral over \partial K is nothing but a sum of integrals over the edges of K with the integrand restricted to each edge seen from the perspective of K.

davelee2804:
Thank you @amartinhuertas ! I had not appreciated the subtlety that in for example Kang et al. eqns (14) and (15) the first skeleton integral is over `\partial K` while the second is over `e`, this makes much more sense now! Looking at (7) in Nguyen et al I see that all the skeleton integrals are over `\partial T`, which is consistent with 
 
> integrand restricted to each edge seen from the perspective of K

as you say. 

I am now implementing a convergence study for the solid body rotation test (second order in space + time), to verify that everything is OK.

A few days ago I got some nonsensical results with order 4, so I will also double check that as well...

amartinhuertas:
> Thank you @amartinhuertas ! I had not appreciated the subtlety that in for example Kang et al. eqns (14) and (15) the first skeleton integral is over \partial K while the second is over e, this makes much more sense now! Looking at (7) in Nguyen et al I see that all the skeleton integrals are over \partial T, which is consistent with

Exactly. You can write hybridizable methods either by edges or grouping together edges into cell boundaries \partial K, for each K. The second approach is the one required to implement static condensation locally at each cell, towards assembly of the global problem on the skeleton, which is the main motivation for hybridizable methods.

> I am now implementing a convergence study for the solid body rotation test (second order in space + time), to verify that everything is OK.
> A few days ago I got some nonsensical results with order 4, so I will also double check that as well...

Ok. Let us see.

davelee2804:
Hey @amartinhuertas , just a heads up that the convergence is sub-optimal. There is too much dispersion, perhaps as a result of how I have formulated the time stepping. I'll take a closer look at this...

amartinhuertas:
>  Hey @amartinhuertas , just a heads up that the convergence is sub-optimal. There is too much dispersion, perhaps as a result of how I have formulated the time stepping. I'll take a closer look at this...

I would try to first solve a steady advection problem with manufactured PDE solution. This way we eliminate noise related to time discretization.

davelee2804:
Thanks @amartinhuertas , yes great point, this is the correct approach, I will do this. 

Just a heads up, I changed the time integration to a stiffly-stable second order implicit runge-kutta scheme, and now the solid body rotation test is converging at the correct rate (second order convergence for second order in space+time discretisation) after a single revolution round the sphere (at least for the low resolutions that I have checked).

davelee2804:
hdg_convergence

davelee2804:
Hi @amartinhuertas , I've added a convergence test for the solid body rotation configuration of the HDG advection solver. I'm wondering if now is a good time to do a code review and merge this branch, before starting work on the linear wave equation with jump penalisation on the pressure gradient tangent component - what do you think?

davelee2804:
...SBRAdvectionHDG tests are failing in github-ci (but passing on my laptop) :
https://github.com/gridapapps/GridapGeosciences.jl/runs/7274708478?check_suite_focus=true#step:9:1037

Should I also commit the Manifest.toml?

amartinhuertas:
> Should I also commit the Manifest.toml?

Yes, because we are pointing to versions of some of the packages which are not yet registered in the Julia registries.

amartinhuertas:
Also, please, pull from `master` if you didnt lately. I see in this PR the following changes that are already in `master`, so that they should not be in this PR. 


https://github.com/gridapapps/GridapGeosciences.jl/pull/34/files#diff-8ad2a593c8bedda9e0bea392ea69c0bbbe41609839d4d3b33bc4a81155a208dcL48

amartinhuertas:
> Hi @amartinhuertas , I've added a convergence test for the solid body rotation configuration of the HDG advection solver. 

I see in the attached plot that there is a mild degradation in slope for the highest resolution that you tested. Have you tried even higher resolutions to be 100% sure that the convergence is actually quadratic?

davelee2804:
> I see in the attached plot that there is a mild degradation in slope for the highest resolution that you tested

Hi @amartinhuertas , higher resolutions are to the left on the x-axis, so the convergence rate is actually increasing as resolution increases. Does that make sense? Have I missed your point somehow?

davelee2804:
...Ie: convergence rate from dx=0.12 to 0.06 is higher than from dx=0.24 to 0.12. Does that make sense?

amartinhuertas:
> ...Ie: convergence rate from dx=0.12 to 0.06 is higher than from dx=0.24 to 0.12. Does that make sense?

Ok, you are right, I was confused. The convergence curve  looks like a piece-wise linear function with 2 pieces, where the right piece seems to be parallel to the theoretical convergence, and the left piece seems to bend down a little bit, with an slope which is no longer parallel to the theoretical convergence. But, as you say, bending down means even faster convergence, and not the opposite. (Here it was my confusion).

amartinhuertas:
> I'm wondering if now is a good time to do a code review and merge this branch

Yes, it might be a good time to do this. The only thing is that I have some other code reviews on the queue, I wont be able to do this immediately.

Apart from this code review, we need:
1) To investigate issue https://github.com/gridapapps/GridapGeosciences.jl/issues/36
2) To fix (if actually required) and test the geometrical discretization of the sphere using an analytical mapping.

davelee2804:
No problem. Note that for the SBRAdvectionHDGTests file that I committed, convergence between the lowest two resolutions is sub-optimal, but quadratic convergence is achieved between the next two.

davelee2804:
No problem @amartinhuertas , there is no rush from me. Please let me know if there is anything I can do to help with any of the issues you mention. We are using an isoparametric mapping right now, is that correct?

amartinhuertas:
> We are using an isoparametric mapping right now, is that correct?

We are using biquadratic elements for the geometry, and bilinear elements for the unknown. I think that the term "isoparametric" is used to refer to those scenarios in which the order of the polynomial to approximate the geometry matches the order of the polynomial to approximate the unknown.

davelee2804:
Aaah, yes, that is what the `2` is for in the CubedSphereDiscreteModel :) OK. Is there a particular reason we need to support an analytical mapping as well?

amartinhuertas:
> Is there a particular reason we need to support an analytical mapping as well?

We have been using an analytical mapping in all the results we obtained with GridapGeosciences + compatible FEs so far. 

If we can represent the geometry exactly, better to stick to it, instead of approximating it, so that we eliminate any noise caused by geometrical errors (leaving integration errors apart). I remember e.g., that, with a bilinear approximation of the cubed sphere we were not able to obtained the theoretical convergence rates of RT+DG FEs for Darcy on the sphere.

davelee2804:
OK, fair enough. To my mind supporting iso or super parametric mappings (as you are already doing) is better, as if we ever want to add mountains when we will have to ditch the analytic mapping anyway, so ensuring everything works for a numerical mapping makes it easier to transition to such a case if we ever want to. But I take your point, analytic is nicer...

amartinhuertas:
Yes, sure, better to have both.

PR #49, opened 2026-05-27
Port to Gridap v0.20 

amartinhuertas:
Serial tests and distributed tests now pass


tamaratambyah:
https://github.com/gridapapps/GridapGeosciences.jl/blob/porting_to_gridap_v0.20/test/Laplacian/Hcurl_MWE.jl

@amartinhuertas I have added a MWE with 6 cells that shows the issue with taking the curl of a Hcurl function

tamaratambyah:
FYI - I updated the MWe to have boundary conditions enforced in the space. But I think the issue still occurs

PR #51, opened 2026-05-29
Add AtlasDiscreteModels: atlas-based discrete models for curved manifolds

santiagobadia:
## Summary

- Introduces `AtlasGrid{Dc,Da}` and `AtlasDiscreteModel{Dc,Da}`, a DG-style grid and discrete model for `Dc`-dimensional manifolds embedded in `Da`-dimensional ambient space, built on a multi-chart atlas.
- Supports serial and distributed (p4est/MPI) execution, intrinsic and extrinsic manifold styles, and analytic metric fields per shape.
- Ships canonical coarse meshes for cylinder, Möbius strip, and cubed sphere, each with explicit analytic metric and inverse-metric fields (avoiding generic JtJ evaluation at assembly time).
- Includes a `lazy_map(Reindex, indices)` specialization that returns `CompressedArray`, ensuring the full pipeline (ambient maps, metrics, gradients, inverse metrics) operates on O(n_charts) values rather than O(n_cells).

## Key files

| File | Contents |
|---|---|
| `AtlasGrids.jl` | `AtlasGrid`, `IdentityField`, `ManifoldStyle`, `JtJ`, `Reindex→CompressedArray` |
| `CoarseMeshes.jl` | `CoarseMesh`, canonical shapes, analytic metric fields, `inverse_metric_field` |
| `AtlasDiscreteModels.jl` | `AtlasDiscreteModel`, visualization, `MetricCellField`, `InvMetricCellField` |
| `DistributedAtlasDiscreteModels.jl` | `AtlasOctreeDistributedDiscreteModel` (p4est) |

## Tests

- Serial geometry: cylinder, Möbius strip, cubed sphere
- Physics: Poisson (extrinsic P2, intrinsic Q2) with O(h³) convergence
- Metric validation: concrete fields agree with JtJ to 1e-12; g·g⁻¹=I
- CompressedArray chain: all lazy arrays have `length(values) == n_charts`
- Distributed: cubed sphere via MPI+p4est
- Physics stubs: Darcy and H(div) on cylinder/sphere

## Test plan

- [x] Run `julia --project=. src/AtlasDiscreteModels/test_metric_fields.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_serial_cylinder.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_serial_mobius.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_serial_cubed_sphere.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_poisson_cylinder.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_poisson_cylinder_intrinsic.jl`
- [x] Run `julia --project=. src/AtlasDiscreteModels/test_compressed_optimization.jl`
- [x] Run `mpiexec -n 4 julia --project=. src/AtlasDiscreteModels/test_distributed_cubed_sphere.jl`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Geometry/AtlasDiscreteModels.jl
```diff
@@ -0,0 +1,284 @@
...
+# Ambient Da-dimensional coordinates are computed HERE (visualization_data only).
+
+include("AtlasGrids.jl")   # also includes CoarseMeshes.jl
+
+# ============================================================
+# AtlasDiscreteModel
+# ============================================================
+
+"""
+    AtlasDiscreteModel{Dc,Da,G,A,P,O,T,L} <: Gridap.Geometry.DiscreteModel{Dc,Dc}
```
amartinhuertas: Why `AtlasDiscreteModel{Dc,Da,G,A,P,O,T,L}` is a subtype of `Gridap.Geometry.DiscreteModel{Dc,Dc}` and NOT `Gridap.Geometry.DiscreteModel{Dc,Da}` ?

src/Geometry/AtlasGrids.jl
```diff
@@ -0,0 +1,434 @@
...
+- (M)                   — `ManifoldStyle` encoded as type parameter; not stored as a field.
+                          Access via `ManifoldStyle(g)`.
+"""
+struct AtlasGrid{Dc, Da,
+                 G <: Gridap.Geometry.Grid{Dc,Dc},
+                 A <: AbstractVector,
+                 P <: AbstractVector{<:Field},
+                 C <: AbstractVector,
+                 O <: Gridap.Geometry.OrientationStyle,
+                 M <: ManifoldStyle} <: Gridap.Geometry.Grid{Dc,Dc}
```
amartinhuertas: see https://github.com/gridapapps/GridapGeosciences.jl/pull/51/changes#r3345260323

amartinhuertas: Ok, I think I now understand why it is currently `Grid{Dc,Dc}` and not `Grid{Dc,Da}`.

`get_cell_coordinates(...)` and `get_node_coordinates(...)` currently return arrays of `Point{Dc}` objects, not `Point{Da}` objects. 

I do not see how `get_cell_coordinates(...)` could return  `Point` objects different from `Point{Dc}`. 

Thus, I think we have the proper sutyping relationship.

amartinhuertas: On the other hand, we have that get_cell_map(...) is fed with Point{Dc} objects, and it returns Point{Da} objects whenever Manifold style is ExtrinsicManifold.  

This sounds like we have an insconsistency here?

Perhaps we should have {Dc,Dc} whenever Manifold style is IntrinsicManifold, and {Dc,Da} whenever Manifold style is ExtrinsicManifold?

In any case, I am not sure how this inconsistency affects the FEM code, etc. Right now it seems we are able to run FEM simulations even with the inconsistency.

src/Geometry/AtlasDiscreteModels.jl
```diff
@@ -0,0 +1,284 @@
...
+and wrap it with the fine model's topology and face labeling.
+
+When `metric_fields` is not provided it is computed lazily from `ambient_maps` via
+`_pullback_metrics(ambient_maps)` for each chart.
+"""
+function AtlasDiscreteModel(
+    coarse_model    :: Gridap.Geometry.DiscreteModel{Dc,Dc},
+    coarse_chart_coords,
+    ambient_maps,
+    num_refinements :: Int = 1;
```
amartinhuertas: For all other outer constructors, `num_refinements` is a mandatory positional argument, while for this one it is an optional positional argument.

Which is the specific reason why this an optional argument? Should we make it mandatory as well for consistency with the other outer constructors?

src/Geometry/AtlasGrids.jl
```diff
@@ -0,0 +1,434 @@
...
+              metric_fields=nothing, orientation_style=nothing, manifold_style=ExtrinsicManifold())
+
+Build an `AtlasGrid` from a coarse model.  When `metric_fields` is not provided it is
+computed lazily via `_pullback_metrics(ambient_maps)` for each ambient map `φ`.
+"""
+function AtlasGrid(
+    coarse_model        :: Gridap.Geometry.DiscreteModel{Dc,Dc},
+    coarse_chart_coords,
+    ambient_maps,
+    num_refinements :: Int = 1;
```
amartinhuertas: see https://github.com/gridapapps/GridapGeosciences.jl/pull/51/changes#r3345511254

src/Geometry/AtlasDiscreteModels.jl
santiagobadia: no specific reason, we can change it

src/Geometry/AtlasGrids.jl
santiagobadia: I think there is no inconsistency. In the intrinsic case Dc = Da, but this is not a problem.

src/Geometry/AtlasGrids.jl
santiagobadia: same

src/Geometry/AtlasDiscreteModels.jl
santiagobadia: I think it should be a subtype of `Gridap.Geometry.DiscreteModel{Dc,Da}`

src/Geometry/AtlasGrids.jl
```diff
@@ -0,0 +1,423 @@
...
+# _lazy_map_compressed when applying gradient/metric/inverse lazily over cells).
+#
+# Candidate for upstreaming to Gridap.Arrays.Reindex.
+function Gridap.Arrays.lazy_map(
+    k       :: Gridap.Arrays.Reindex{<:AbstractArray},
+    ::Type{T},
+    j_to_i  :: AbstractArray,
+) where T
+  Gridap.Arrays.CompressedArray(k.values, j_to_i)
+end
```
santiagobadia, line 100: We should probably add an if such that we only create the CompressedArray if j_to_i has nothing as cache

src/Geometry/AtlasGrids.jl
amartinhuertas: > I think there is no inconsistency. In the intrinsic case Dc = Da, but this is not a problem.

I agree there is no inconsistency whenever we have `{Dc,Da}`. My comments were referring to the case in which we had `<:DiscreteModel{Dc,Dc}`.

The only inconsistency that we still have even if we put <:DiscreteModel{Dc,Da}, is that get_cell_coordinates(...) and get_node_coordinates(...) are not quite correct whenever Da is different from Dc. But I do no think we can do much about it.

src/Geometry/AtlasDiscreteModels.jl
amartinhuertas: > I think it should be a subtype of Gridap.Geometry.DiscreteModel{Dc,Da}

Ok, agreed.

src/Geometry/AtlasGrids.jl
amartinhuertas: Done

src/Geometry/AtlasDiscreteModels.jl
amartinhuertas: Done

src/Geometry/AtlasGrids.jl
amartinhuertas: Done

src/Geometry/AtlasDiscreteModels.jl
amartinhuertas: Done.

PR #54, opened 2026-06-10
Testing the latest version of dev gridap master to see whether  we can run Hcurl-conforming simulations with change_dof=true, i.e., the current default option

amartinhuertas:


amartinhuertas:
I finally checked this manually in my local computer, and I can confirm that the fix worked.

